PR #1, opened 2021-03-07
change marginToken use to pool.amount in insurance

raymogg:
# Motivation
There was a potential exploit in the drainPool code that could be used by malicious Tracer markets to drain all of a single token from insurance (all the DAI, not just DAI from an individual pool). This was due to the fact that the drainPool function was referencing the balance of the marginToken for the entire insurance contract, and not just the individual market.

# Changes
- changed references in the drainPool contract from the balance of the insurance contract itself to the pool balance. This way draining is capped to the balance of the pool


CalabashSquash (approved):
:cactus:

PR #5, opened 2021-03-09
Add a slippage cap that limits the amount of slippage a liquidator ca…

CalabashSquash:
### Motivation
We want a stop-gap measure to limit the amount of slippage that a liquidator can claim.

### Changes
- Added governance ownership to the Account.sol and Receipt.sol contracts, so contract variables can be set/modified
- Added a `maxSlippage` contract variable and `setMaxSlippage` to Receipt.sol, which indicates that maximum percent slippage multiplied by 10,000 that a liquidator can claim.
- Modified `Receipt.claimReceipts` to enforce `maxSlippage`. If the slippage is greater than `maxSlippage`, set slippage to `maxSlippage`.
- Added `setMaxSlippage` to Receipt.
- Added test to make sure slippage is appropriately capped.

### DIFF
--- a/contracts/Receipt.sol
+++ b/contracts/Receipt.sol
@@ -3,6 +3,7 @@ pragma solidity >=0.6.0;
 pragma experimental ABIEncoderV2;
 import "@openzeppelin/contracts/math/SafeMath.sol";
 import "@openzeppelin/contracts/math/SignedSafeMath.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
 import {Types} from "./Interfaces/Types.sol";
 import "./lib/LibMath.sol";
 import "./Interfaces/IReceipt.sol";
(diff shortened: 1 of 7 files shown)


contracts/Receipt.sol
```diff
@@ -13,13 +14,14 @@ import "./Interfaces/ITracer.sol";
...
 */
-contract Receipt is IReceipt {
+contract Receipt is IReceipt, Ownable {
     using SafeMath for uint256;
     using SignedSafeMath for int256;
     using LibMath for uint256;
     using LibMath for int256;
 
     uint256 public override currentLiquidationId;
+    int256 public override maxSlippage;
```
sporejack, line 24: Is there an explicit reason this is signed?

contracts/Receipt.sol
```diff
@@ -230,6 +246,10 @@ contract Receipt is IReceipt {
         );
     }
 
+    function setMaxSlippage(int256 _maxSlippage) public override onlyOwner() {
```
sporejack, line 249: See previous comments on signedness of `_maxSlippage`.

raymogg (approved):

sporejack (changes requested):

contracts/Receipt.sol
CalabashSquash: We're gonna have to convert it to int256 for calculations regardless, and it doesn't ever need to be very large.

PR #6, opened 2021-03-09
Update Funding Rate Sensitivity Setter

raymogg:
# Motivation
Previously the funding rate sensitivity was a constant 1 per market. This is now settable on deploy of the market, and also can be changed by the owner of a market at any point in time. This PR also removes the redundant `minMargin` param from deployment of markets.

# Changes
- remove `minMargin`
- add `fundingRateSensitivity` to deployment
- add a setter for funding rate sensitivity


CalabashSquash (approved):
:shipit:

PR #8, opened 2021-03-10
Deploy insurance pool in TracerFactory.deployTracer

CalabashSquash:
#### Insurance.sol
- Removed `Insurance.deployInsurancePool`'s `onlyOwner` modifer. We want anyone to be able to deploy an insurance pool of a Tracer
- Added a check in `Insurance.deployInsurancePool` to make sure the market is a valid Tracer according to the factory.
    - For the above to work we need to be able to have a factory contract variable to check. This was added.
    - **This needs to be set after insurance contract is deployed**

#### Receipt.sol
- Unused variables

#### Tracer.sol
- Made `insuranceContract` a `IInsurance`, not just `address`.

#### TracerFactory.sol
- Calls `Insurance.deployInsurancePool` during `deployTracer`.

#### Migration script
- The insurance contract was already set in constructor, so doesn't need to be set again via proposal. This proposal was removed.

### DIFF
--- a/contracts/Receipt.sol
+++ b/contracts/Receipt.sol
@@ -93,7 +93,7 @@ contract Receipt is IReceipt, Ownable {
         require(block.timestamp < receipt.releaseTime, "REC: claim time passed");
         require(!receipt.liquidatorRefundClaimed, "REC: Already claimed");
         // Validate the escrowed order was fully sold
-        (uint256 unitsSold, int256 avgPrice) = calcUnitsSold(orderIds, escrowId, market, receipt.time);
+        (uint256 unitsSold, int256 avgPrice) = calcUnitsSold(orderIds, escrowId, market);
         require(
             unitsSold == uint256(receipt.amountLiquidated.abs()),
             "REC: Unit mismatch"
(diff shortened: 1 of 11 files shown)


contracts/Receipt.sol
```diff
@@ -93,7 +93,7 @@ contract Receipt is IReceipt, Ownable {
         require(block.timestamp < receipt.releaseTime, "REC: claim time passed");
         require(!receipt.liquidatorRefundClaimed, "REC: Already claimed");
         // Validate the escrowed order was fully sold
-        (uint256 unitsSold, int256 avgPrice) = calcUnitsSold(orderIds, escrowId, market, receipt.time);
```
raymogg, line 96: wasn't mentioned in the PR, just wondering why `receipt.time` was removed?

raymogg (changes requested):

contracts/Receipt.sol
CalabashSquash: It was unused, because I just use receipt.time inside `calcUnitsSold`, so it's unnecessary

raymogg (approved):

PR #9, opened 2021-03-10
Fix contradictory comparison in TWAP code

sporejack:
# Description #
The TWAP calculation code in `Pricing.sol` contained an always-false comparison, where an unsigned integer was checked for negativity. This PR fixes this redundancy and simplifies some of the TWAP calculation logic.

### DIFF
--- a/contracts/Pricing.sol
+++ b/contracts/Pricing.sol
@@ -272,8 +272,8 @@ contract Pricing is IPricing {
             if (j < 0) {
                 j = 23;
             }
-            int256 derivativePrice = getHourlyAvgTracerPrice(j, market);
-            int256 underlyingPrice = getHourlyAvgOraclePrice(j, market);
+            int256 derivativePrice = getHourlyAvgTracerPrice(uint256(j), market);
+            int256 underlyingPrice = getHourlyAvgOraclePrice(uint256(j), market);
             if (derivativePrice != 0) {
                 derivativeInstances = derivativeInstances.add(uint256(timeWeight));
                 derivativeSum = derivativeSum.add((timeWeight).mul(derivativePrice));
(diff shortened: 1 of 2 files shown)


contracts/Pricing.sol
```diff
@@ -267,10 +267,10 @@ contract Pricing is IPricing {
         uint256 underlyingInstances = 0;
         for (uint8 i = 0; i < 8; i++) {
             int256 timeWeight = 8 - i;
-            uint256 j = currentHour - i; // keep moving towards 0
+            int256 j = int256(currentHour) - int256(i); // keep moving towards 0
             // loop back around list if required
             if (j < 0) {
-                j = 24 + (currentHour - i);
+                j = 23;
```
CalabashSquash, line 273: If currentHour != i - 1, j might not necessarily equal 23.

On phone at the moment, so can't check properly for another couple of hours, but just wanted to make sure this was known.

raymogg: @BenjaminPatch Jack and I whiteboarded this yesterday arvo and the conclusion we came to was that currentHour is in [0,23]. And in this snippet the `if j < 0` case should always be called when j = -1, since j will start at current hour and move towards 0.

It seemed to us that having 24 + (currentHour - i) then meant that since this always gets triggered when j (which is equal to currentHour - 1 anyway) is equal to -1, then this should always be a constant 23. Thoughts?

raymogg (changes requested):
@sporejack what was the reason for changing all the hour params to a int256 over uint256?

Other than that LGTM in terms of logic of changes

sporejack:
> @sporejack what was the reason for changing all the hour params to a int256 over uint256?
> 
> Other than that LGTM in terms of logic of changes

Not changing them would necessitate casting `j` from a signed type to an unsigned type at the call site for the relevant functions.

Prior to the simplifications we made (i.e., just setting `j` to be 23), it was less clear whether this was meaningful or not. In hindsight, it's fine provided our (admittedly fragile) invariant of `j != -1` is preserved by the time it comes to cast `j`.

raymogg (approved):

PR #11, opened 2021-03-11
Alter off chain trade execution flow

raymogg:
# Motivation
This branch contains updated `trader.sol` code to improve the way off chain orders. are handled on chain. To do this, the simple dex and tracer contracts are also updated

# Changes
- Change the executeTrades function in the trader contract to match 1-1
- introduce a matchOrders function in tracer.sol
- introduce a mapping of order hash to id in simple dex

### DIFF
--- a/contracts/DEX/SimpleDex.sol
+++ b/contracts/DEX/SimpleDex.sol
@@ -1,26 +1,27 @@
 //SPDX-License-Identifier: GPL-3.0-or-later
-pragma solidity >=0.6.0 <0.7.0;
+pragma solidity >0.6.0;
 pragma experimental ABIEncoderV2;
 
-import "@openzeppelin/contracts/math/SafeMath.sol";
-import "@openzeppelin/contracts/math/SignedSafeMath.sol";
-import "../lib/LibMath.sol";
-import "../Interfaces/Types.sol";
+import '@openzeppelin/contracts/math/SafeMath.sol';
+import '@openzeppelin/contracts/math/SignedSafeMath.sol';
+import '../lib/LibMath.sol';
+import '../Interfaces/Types.sol';
+import '../Interfaces/IDex.sol';
 
 /**
  * SimpleDex Contract: Implements the Tracer make/take without underlying
  * management checks.
  */
-contract SimpleDex {
+contract SimpleDex is IDex {
     using SafeMath for uint256;
     using SignedSafeMath for int256;
     using LibMath for uint256;
     using LibMath for int256;
 
-    //State variables
-    uint256 public orderCounter = 0;
+    // Order counter starts at 1 due to logic in Trader.sol
+    uint256 public orderCounter = 1;
     mapping(uint256 => Types.Order) public orders;
-
+    mapping(bytes32 => uint256) public override orderIdByHash;
 
     /**
      * @notice Places an on chain order via a permissioned contract, fillable by any part on chain.
@@ -38,7 +39,15 @@ contract SimpleDex {
         address maker
     ) internal returns (uint256) {
         //Create unfilled order
-        orders[orderCounter] = Types.Order(maker, amount, price, 0, side, expiration, block.timestamp);
+        Types.Order storage order = orders[orderCounter];
+        order.maker = maker;
+        order.amount = amount;
+        order.price = price;
+        order.side = side;
+        order.expiration = expiration;
+        order.creation = block.timestamp;
+        //Map order hash to id
+        orderIdByHash[hashOrder(amount, price, side, expiration, maker)] = orderCounter; 
         orderCounter++;
         return orderCounter - 1;
     }
@@ -54,12 +63,20 @@ contract SimpleDex {
         uint256 orderId,
         uint256 amount,
         address _taker
-    ) internal returns (Types.Order memory, uint256, uint256, address) {
+    )
+        internal
+        returns (
+            Types.Order memory,
+            uint256,
+            uint256,
+            address
+        )
+    {
         //Fill or partially fill order
         Types.Order storage order = orders[orderId];
-        require(order.amount.sub(order.filled) > 0, "SDX: Order filled");
+        require(order.amount.sub(order.filled) > 0, 'SDX: Order filled');
         /* solium-disable-next-line */
-        require(block.timestamp < order.expiration, "SDX: Order expired");
+        require(block.timestamp < order.expiration, 'SDX: Order expired');
 
         //Calculate the amount to fill
         uint256 fillAmount = (amount > order.amount.sub(order.filled)) ? order.amount.sub(order.filled) : amount;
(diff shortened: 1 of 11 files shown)


contracts/DEX/SimpleDex.sol
```diff
@@ -38,7 +39,15 @@ contract SimpleDex {
...
     ) internal returns (uint256) {
         //Create unfilled order
-        orders[orderCounter] = Types.Order(maker, amount, price, 0, side, expiration, block.timestamp);
+        Types.Order storage order = orders[orderCounter];
+        order.maker = maker;
+        order.amount = amount;
+        order.price = price;
+        order.side = side;
+        order.expiration = expiration;
+        order.creation = block.timestamp;
```
CalabashSquash, line 48: Was this change for forward compatibility with solc 0.7?

contracts/DEX/SimpleDex.sol
```diff
@@ -54,12 +63,20 @@ contract SimpleDex {
...
+            address
+        )
+    {
         //Fill or partially fill order
         Types.Order storage order = orders[orderId];
-        require(order.amount.sub(order.filled) > 0, "SDX: Order filled");
+        require(order.amount.sub(order.filled) > 0, 'SDX: Order filled');
         /* solium-disable-next-line */
-        require(block.timestamp < order.expiration, "SDX: Order expired");
+        require(block.timestamp < order.expiration, 'SDX: Order expired');
```
CalabashSquash: Why change from double quotes to single quotes?

contracts/DEX/SimpleDex.sol
```diff
@@ -71,4 +88,61 @@ contract SimpleDex {
...
+        // Fill or partially fill order
+        Types.Order storage order1 = orders[order1Id];
+        Types.Order storage order2 = orders[order2Id];
+
+        // Ensure orders can be cancelled against each other
+        require(order1.price == order2.price, 'SDX: price mismatch');
+        
+        /* solium-disable-next-line */
+        require(block.timestamp < order1.expiration &&
+            block.timestamp < order2.expiration, 'SDX: Order expired');
```
CalabashSquash: More single quotes

contracts/DEX/SimpleDex.sol
```diff
@@ -71,4 +88,61 @@ contract SimpleDex {
...
+
+        // Calculate the amount to fill
+        uint256 order1Remaining = order1.amount.sub(order1.filled);
+        uint256 order2Remaining = order2.amount.sub(order2.filled);
+        uint256 fillAmount = 0;
+        if (order1Remaining >= order2Remaining) {
+            // Use order2Remaining amount
+            fillAmount = order2Remaining;
+        } else {
+            // Use order1Remaining amount
```
CalabashSquash: These comments seem redundant

contracts/DEX/SimpleDex.sol
```diff
@@ -71,4 +88,61 @@ contract SimpleDex {
...
-}
+
+    /**
+    * @notice Matches two orders that have already both been made. Has the same
+    *         validation as takeOrder 
+    */
+    function _matchOrder(
+        uint256 order1Id,
+        uint256 order2Id
+    ) internal returns (uint256) {
```
CalabashSquash, line 99: So is this function essentially making it so two orders created from `makeOrder`, can both be filled against eachother?

contracts/Tracer.sol
```diff
@@ -251,6 +251,75 @@ contract Tracer is ITracer, SimpleDex, Ownable {
...
+    * @param order2 the second order that exists on chain
+    */
+    function matchOrders(uint order1, uint order2) public override {
+        uint256 fillAmount = _matchOrder(order1, order2);
+
+        int256 orderPrice = orders[order1].price;
+        address order1User = orders[order1].maker;
+        address order2User = orders[order2].maker;
+        bool order1Side = orders[order1].side;
+        int256 baseChange = (fillAmount.mul(uint256(orderPrice))).div(priceMultiplier).toInt256();
```
CalabashSquash, line 267: I was going to write a comment about how there should be a check to make sure prices are the same, but then realised it is done in `SimpleDex._matchOrder`. I feel like there should be a comment above line 260 saying `// Perform compatibility checks and calculate fill amount` or something.

On the same now, should there be a check to make sure each order has different sides? At the moment it just checks one and assumes the other is the opposite.

contracts/Trader.sol
```diff
@@ -45,63 +46,32 @@ contract Trader {
...
-                currentMake.side,
-                currentMake.expiration,
-                currentMake.user
-            );
+        require(makers.length == takers.length, 'TDR: Lengths differ');
+
+        // safe as we've already bounds checked the array lengths
+        uint256 n = makers.length;
+
+        require(n > 0, 'TDR: Received empty arrays');
```
CalabashSquash: More single quotes

contracts/Trader.sol
```diff
@@ -112,39 +82,51 @@ contract Trader {
...
-                ),
-                "TDR: incorrect order sig or nonce"
-            );
+    function grabOrder(Types.SignedLimitOrder[] memory orders, uint256 index, address market)
+        internal
+        returns (uint256)
+    {
+        require(index <= orders.length, 'TDR: Out of bounds access');
+
+        IDex dex = IDex(market);
```
CalabashSquash, line 91: Is this the address of the Tracer market? If so, how does it work? I thought it would only work if casting from an instance of IDex interface, whereas Tracer is ITracer.

contracts/Trader.sol
```diff
@@ -112,39 +82,51 @@ contract Trader {
...
+            signedOrder.sigR,
+            signedOrder.sigS,
+            signedOrder.sigV
+        );
 
-            return currentSigned.order;
+        bytes32 orderHash = hashOrderForDex(signedOrder.order);
+        // check if order exists on chain, if not, create it
+        uint orderId = dex.orderIdByHash(orderHash);
+        if (orderId == 0) {
```
CalabashSquash, line 107: What about the first order?

CalabashSquash: Nevermind, saw it was changed to 1 as first

CalabashSquash (changes requested):

contracts/Tracer.sol
raymogg: Agreed on the comment.  I want to do another critical sweep and really think where all the logic is sitting. What sits in Trader vs what sits in Tracer / SimpleDex.

Ah yes there should most definitely be a side. check.

contracts/DEX/SimpleDex.sol
raymogg: Yep thats it. This could even become a keeper type roll, emptying the order book

contracts/Trader.sol
raymogg: Yeah it is. Tracer also implements the IDex interface because of the line `Tracer is SimpleDex`. In this way, the SimpleDex interface is actually implemeneted by Tracer itself

contracts/DEX/SimpleDex.sol
raymogg: I was running it on Solidity 0.6.0, I'll find us the latest stable OZ version to run and get a ticket up for that migration

contracts/Trader.sol
CalabashSquash: Ah true, I didn't realise.

contracts/Tracer.sol
```diff
@@ -257,6 +257,7 @@ contract Tracer is ITracer, SimpleDex, Ownable {
     * @param order2 the second order that exists on chain
     */
     function matchOrders(uint order1, uint order2) public override {
+        // perform compatability checks (price, side) and calc fill amount
```
CalabashSquash: compatability -> compatibility

CalabashSquash (approved):

contracts/DEX/SimpleDex.sol
```diff
@@ -71,4 +88,59 @@ contract SimpleDex {
...
+
+        // Fill or partially fill order
+        Types.Order storage order1 = orders[order1Id];
+        Types.Order storage order2 = orders[order2Id];
+
+        // Ensure orders can be cancelled against each other
+        require(order1.price == order2.price, "SDX: Price mismatch");
+
+        // Ensure orders are for opposite sides
+        require(order1.side != order2.side, "SDX: Side mismatch");
```
sporejack: Reversion message is slightly misleading as a "mismatch" implies that the sides *don't* match (which is exactly what we want).

sporejack (changes requested):
LGTM -- apart from one nitpick.

contracts/DEX/SimpleDex.sol
raymogg: @sporejack how about "SDX: Same side"?

PR #12, opened 2021-03-12
Removed deprecated calculations. Fixed Trader.sol tests

CalabashSquash:
### Motivation
Functions that use the calculations from the old accounting system were still in the codebase. These should be deleted.

Also, a test in the Trader.sol unit tests was passing incorrectly 

### Changes
- Deleted `LibBalance.sol`'s safeCalcMarginPercent, calcMarginPercent, calcPositiveNegative
    - Changed tests using these calculations to the new system.
    - Deleted `Account.unsafeGetUserMargin` 
- Increment nonce on order making in Trader.sol
- Changed tests in unit/Trader.ts
    - Changed order sides to be opposite of order being matched against
    - Added a test for when order is made twice (invalid)
    - Check that `Trader.executeTrade` is actually incrementing order nonce.

### DIFF
--- a/test-ts/unit/Trader.ts
+++ b/test-ts/unit/Trader.ts
@@ -2,7 +2,7 @@
 import { BN, expectRevert, time } from "@openzeppelin/test-helpers"
 import assert from "assert"
 import { setupContractsAndTracer } from "../lib/Setup"
-import { AccountInstance, TracerInstance, TraderInstance } from "../../types/truffle-contracts"
+import { AccountInstance, TracerInstance, TraderInstance, TestTokenInstance } from "../../types/truffle-contracts"
 import { signOrder, signOrders, domain, domainData, limitOrder } from "../lib/Signing"
 import { accounts, configure, web3 } from "../configure"
 import { Trader } from "../artifacts"
@@ -10,6 +10,8 @@ import { Trader } from "../artifacts"
 describe("Trader Shim unit tests", async () => {
     let trader: TraderInstance;
     let tracer: TracerInstance;
+    let account: AccountInstance;
+    let token: TestTokenInstance;
     
     let sampleMakers: any;
     let sampleTakers: any;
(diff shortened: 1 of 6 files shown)


test-ts/unit/Trader.ts
```diff
@@ -102,6 +108,24 @@ describe("Trader Shim unit tests", async () => {
     })
 
     describe("executeTrade", () => {
+        context("When order is executed twice", () => {
+            it("passes", async () => {
```
sporejack: Should be "reverts"?

test-ts/unit/Trader.ts
```diff
@@ -102,6 +108,24 @@ describe("Trader Shim unit tests", async () => {
...
+                /* sign orders for submission */
+                let signedMakers: any = await Promise.all(await signOrders(web3, makers, trader.address));
+                let signedTakers: any = await Promise.all(await signOrders(web3, takers, trader.address));
+
+                assert(await trader.executeTrade(signedMakers, signedTakers, market));
+                await expectRevert(
+                    trader.executeTrade(signedMakers, signedTakers, market),
+                    "TDR: Incorrect nonce"
+                );
+            })
```
sporejack: Additionally, I think there should be another `it` call which checks that the nonces were actually updated correctly (i.e., incremented).

test-ts/unit/Trader.ts
```diff
@@ -141,7 +165,10 @@ describe("Trader Shim unit tests", async () => {
                 let signedMakers: any = await Promise.all(await signOrders(web3, makers, trader.address));
                 let signedTakers: any = await Promise.all(await signOrders(web3, takers, trader.address));
 
-                assert(trader.executeTrade(signedMakers, signedTakers, market));
+                assert(await trader.executeTrade(signedMakers, signedTakers, market));
```
sporejack, line 150: Is this block misplaced?

sporejack (changes requested):
The Trader shim modifications need more attention, but the other aspects of the PR are fine.

sporejack (approved):

PR #16, opened 2021-03-14
depositTo Patch

raymogg:
# motivation
The deposit to function was accepting a depositer param, however this should have just used `msg.sender` to enforce the person calling the function to deposit into another users account.

# changes
- remove depositer param
- use msg.sender


CalabashSquash (approved):
:surfing_man:

PR #17, opened 2021-03-15
Add percentage factor offset to funding rate sensitivity

kumar-ish:
# Motivation
Lets funding rate sensitivity support up to 5 decimal places by adding a factor offset of 10^5. The funding rate sensitivity can now be as low as 0.00001 instead of the minimum of 1 previously. 

# Changes
- Offset funding rate sensitivity by percent factor (10^5)
- Change funding rate sensitivity in `Setup.ts` to be increased by new percent factor

### DIFF
--- a/contracts/Pricing.sol
+++ b/contracts/Pricing.sol
@@ -12,6 +12,7 @@ import "@openzeppelin/contracts/math/SignedSafeMath.sol";
 
 contract Pricing is IPricing {
     uint256 private constant DIVIDE_PRECISION = 100000000; // 10^7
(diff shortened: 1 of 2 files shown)


CalabashSquash (commented):
There should be tests for funding rate sensitivity imo

kumar-ish:
Can do. Just for future reference, the funding rate factor offset is arbitrary and should be increased in the future (at least by a magnitude of 3).

raymogg:
Agreed with @BenjaminPatch. Should be a test that tests a decimal funding rate works as expected imo. Otherwise LGTM

PR #20, opened 2021-03-19
Use tracer utils

dospore:
- Abstract some of the signing logic out so it can be used elsewhere
- Import signing logic from tracer-utils
- Renamed tracer package to tracer-protocol since npm said that tracer was taken :(


migrations-ts/2_deploy_contracts.ts
```diff
@@ -48,15 +48,16 @@ try {
...
 
     console.log("Writing out contract addresses")
     const obj = {
         'account': account,
         'factory': factory,
         'insurance': insurance,
         'pricing': pricing,
-        'oracle': oracle
+        'oracle': oracle,
+	'trader': trader
```
sporejack, line 60: Indentation.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@tracer/contracts",
-  "version": "0.1.0",
+  "name": "@tracer-protocol/contracts",
+  "version": "0.0.1",
```
sporejack, line 3: Should probably reach consensus on versioning here.

sporejack (changes requested):
LGTM -- minor changes though.

raymogg:
Closed due to inactivity. We can use `tracer-utils` at another point.

PR #23, opened 2021-03-25
Safety withdrawals

CalabashSquash:
### Motivation
The insurance pool cannot currently deal with tokens being sent directly to it. Users must deposit, or it must draw from its balance in the account contract.

It could be worth setting safety features in place to recover funds accidentally sent to the insurance pool and other contracts following a similar pattern.
### Changes
- Created `contracts/lib/SafetyWithdraw.sol`, containing `withdrawERC20` to withdraw any ERC20 token from the contract.
- Imported this into Account.sol, Insurance.sol, InsurancePoolToken.sol, Tracer.sol


package.json
```diff
@@ -48,7 +48,7 @@
     "web3-eth-contract": "^1.3.0"
   },
   "scripts": {
-    "test": "cross-env NODE_OPTIONS=--max-old-space-size=4096 mocha test",
+    "test": "cross-env NODE_OPTIONS=--max-old-space-size=4096 mocha test/test-ts/unit/Account.js",
```
raymogg: KILL IT!

contracts/lib/SafetyWithdraw.sol
```diff
@@ -0,0 +1,16 @@
...
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+contract SafetyWithdraw is Ownable, ISafetyWithdraw {
+
+    function withdrawERC20Token(
+        address tokenAddress,
+        address to,
+        uint256 amount
+    ) external override onlyOwner {
+        IERC20(tokenAddress).transfer(to, amount);
```
raymogg, line 14: I worry this can be seen as a rug (owner can withdraw any tokens). I guess it has to go through a gov proposal so its not exactly a "rug".

contracts/Tracer.sol
```diff
@@ -1,22 +1,23 @@
...
 import {Types} from "./Interfaces/Types.sol";
-import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
 import "./Interfaces/IOracle.sol";
 import "./Interfaces/IInsurance.sol";
 import "./Interfaces/IAccount.sol";
 import "./Interfaces/ITracer.sol";
 import "./Interfaces/IPricing.sol";
 import "./DEX/SimpleDex.sol";
+import "@openzeppelin/contracts/math/SafeMath.sol";
```
raymogg, line 14: Just double checking, all these extra imports needed?

raymogg (changes requested):

contracts/Tracer.sol
sporejack: Not extras I don't think?

CalabashSquash: They're just shuffled around so yeah. Just reordering to make local contracts on top, and external ones beneath.

contracts/lib/SafetyWithdraw.sol
CalabashSquash: Should we do anything about this?

sporejack (changes requested):
LGTM - only exception is the test runner being changed in the package manifest.

On the topic of (perceived) rugpulling, provided this has to go via the usual governance system then I believe this is a moot point.

raymogg (approved):

raymogg:
Keeping note of the rugpull argument and will brainstorm if there is a way to do this in a way that is trustless to that (maybe by ensuring tokens being withdrawn aren't actually in use by insurance or somthing?). Might make the emergency withdraw more expensive, but cause less centralisation concern

jmcph4 (approved):

sporejack (approved):

PR #24, opened 2021-03-25
Removed all references to Types.AccountBalance.deposited

### LINKED ISSUE #26: Remove Types.UserBalance.deposited
### Problem
The deposited field is not actually being used, and creates a bug in [Account.sol](https://github.com/tracer-protocol/tracer-protocol/blob/master/contracts/Account.sol#L138) causing nobody to be able to withdraw more than they deposited (profits).

### Solution
Remove all references to Types.UserBalance.deposited

CalabashSquash:
### Motivation
We don't use `Types.AccountBalance.deposited` anymore. And it actually creates a bug in `withdraw` where you can never withdraw more than you deposit (i.e. can't withdraw profit)
### Changes
- Removed any reference to `AccountBalance.deposited`

Closes #26


raymogg:
@dospore does removing deposited affect anything from a UI perspective? Thought it may be worth bringing up

raymogg (approved):

PR #25, opened 2021-03-25
Custom migration network

CalabashSquash:
### Motivation
We want an easy way to tell the migration script "I want to run the migrations with the intent of testing the OME".
### Changes
- Added network "ome" in `truffle-config.js`, which is the same as `development`, but the new name allows the migration script to differentiate between the two.
- Tidied up migration script a bit
- Now setups up the accounts, as it was doing for Kovan before, on `ome` network.
- Tidy up error message for when `*.secret` files not found.


migrations-ts/2_deploy_contracts.ts
```diff
@@ -162,8 +170,8 @@ async function setupSingleAccount(
     }
     // Transfer each address a bit of ETH
     // Then transfer tokens from token deployer address to new address
-    if (network == "development") {
-        // Since we are on development(localhost), we just want to send transactions direclty from
+    if (network == "ome") {
+        // Since we are on network "ome"(localhost), we just want to send transactions direclty from
```
sporejack, line 174: Typo: `s/direclty/directly/gI`.

sporejack (changes requested):
This looks really nice and clean: only flagged a typo in a comment.

CalabashSquash (commented):
Approved @raymogg , but I can't approve officially due to me being the author

PR #28, opened 2021-04-06
Trader gas accounting

sporejack:
# Motivation #
For **off-chain trading** there's currently no way of accounting for and enforcing use of gas on the part of traders. As such, executioner operators would have to foot the bill for all gas spent by their executioner nodes. This is clearly unsustainable (and requires said operators to operate as charities, essentially). This PR remedies this situation.

# Changes #
 - Added a public mapping to the Trader shim to store gas balances **per user**
 - Added gas balance checks to trade execution
 - Added the ability to deposit gas into the Trader shim contract itself
 - Added the ability for the DAO to withdraw ETH from the Trader shim contract


contracts/Trader.sol
```diff
@@ -63,6 +66,9 @@ contract Trader {
 
         require(n > 0, "TDR: Received empty arrays");
 
+        // gas cost for individual traders
+        uint256 gasCost = (tx.gasprice) / (2 * n);
```
raymogg: think this should be `(tx.gasprice * tx.gasused) / (2 * n)`. I assume gas price is just the gas price in wei?

raymogg (changes requested):

contracts/Trader.sol
sporejack: Good catch!

sporejack: @raymogg Consult 6538408.

contracts/Trader.sol
```diff
@@ -67,7 +67,7 @@ contract Trader is Ownable {
         require(n > 0, "TDR: Received empty arrays");
 
         // gas cost for individual traders
-        uint256 gasCost = (tx.gasprice * tx.gasused) / (2 * n);
+        uint256 gasCost = (tx.gasprice * gasleft()) / (2 * n);
```
raymogg, line 70: @sporejack this would mean we are subsidising all gas up to this point. Should this be `gasused + gasleft()`?

raymogg:
Migrated these changes to branch that had updated hardhat setup for ease of migration. New PR opened

PR #31, opened 2021-04-12
Trader nonce validation

sporejack:
# Motivation
Currently, the Trader shim (in `Trader.sol`) increments account nonces even if an order is not fully matched. This obviously breaks partial order functionality because an order can only have one nonce, which will be invalidated the first time the Trader shim sees it.

# Changes
 - Only bump nonces on complete matchings


CalabashSquash (approved):
:womans_hat:

raymogg (changes requested):
Would be good to get tests for this baked into Trader unit tests

raymogg (approved):

PR #33, opened 2021-04-16
Rename tracer tracerperps

CalabashSquash:
### Motivation
Tracer.sol is only perpetual swaps, so to name our perpetual swaps the same as the entire protocol is overly-ambiguous.

### Changes
- Renamed all instances of `Tracer.sol` to `TracerPerpetualSwaps.sol` and `TracerFactory.sol` to `TracerPerpetualsFactory.sol`
- Went through all files to make these changes functional


raymogg (approved):

contracts/Pricing.sol
```diff
@@ -17,7 +17,7 @@ contract Pricing is IPricing {
     using LibMath for uint256;
     using LibMath for int256;
 
-    ITracerFactory public factory;
+    ITracerPerpetualsFactory public factory;
```
raymogg: Do we care that sometimes this is referenced as `factory` and other times its referenced as `perpsFactory`?

CalabashSquash: Went through and changed them all

PR #34, opened 2021-04-18
Calculate max leverage

CalabashSquash:
### Motivation
Deleveraging is a desired feature for when insurance pool starts draining.

### Changes
**TracerPerpetualSwaps.sol**
- Added `int256 INSURANCE_DELEVERAGING_CLIFF`: the upper threshold for the insurance pool when deleveraging begins
    - This is a market parameter.
- Moved `settle` to be above account updates in `takeOrder` and `matchOrders`.

**Account.sol**
- Added function `realMaxLeverage`, which adjusts the max leverage as insurance pool slides from `TracerPerpetualSwaps.INSURANCE_DELEVERAGING_CLIFF`% of target to 0% of target.

### Justification for choosing e.g. 20 as INSURANCE_DELEVERAGING_CLIFF instead of mulitplying by 10,000
- I just figured there is no benefit in having any more accuracy beyond 1 percent.


sporejack (approved):
LGTM!

contracts/Account.sol
```diff
@@ -67,6 +66,44 @@ contract Account is IAccount, Ownable, SafetyWithdraw {
...
+        //int256 oneXLeverage = 10000;
+        // Linear function intercepting points (1, 0) and (INSURANCE_DELEVERAGING_CLIFF, baseMaxLeverage)
+        // y = mx + b,
+        // where m = INSURANCE_DELEVERAGING_CLIFF - 1
+        //       x = percentFull
+        //       b = 1 (since 1 is the y-intercept)
+        // 1*10,000
+
+        int256 one = PERCENT_PRECISION;
+        int256 realMaxLeverage =
```
raymogg: I'm struggling to understand this maths here. So we take the ratio of baseMaxLeverage to the cliff (eg baseMaxLeverage = 100x, cliff = 10%). We then use this ratio and multiply by how full the pool is to produce the realMaxLeverage?

contracts/Account.sol
```diff
@@ -67,6 +66,44 @@ contract Account is IAccount, Ownable, SafetyWithdraw {
...
+            .div(insurance.getPoolTarget(market).toInt256()).mul(100);
+        int256 cliff = _tracer.INSURANCE_DELEVERAGING_CLIFF();
+
+        if (percentFull > cliff.mul(PERCENT_PRECISION)) {
+            return baseMaxLeverage;
+        }
+
+        // Use int256 for compatibility with baseMaxLeverage
+        //int256 oneXLeverage = 10000;
+        // Linear function intercepting points (1, 0) and (INSURANCE_DELEVERAGING_CLIFF, baseMaxLeverage)
```
raymogg: does this imply at percentFull = 1%, the real max leverage is 0?.... so nobody can take any positions if the pool is less than 1% full?

contracts/TracerPerpetualSwaps.sol
```diff
@@ -198,17 +202,17 @@ contract TracerPerpetualSwaps is ITracerPerpetualSwaps, SimpleDex, Ownable, Safe
 
         int256 baseChange = (fillAmount.toInt256().mul(order.price)).div(priceMultiplier.toInt256());
         require(baseChange > 0, "TCR: Base change <= 0");
+        
+        // Settle accounts
```
raymogg, line 207: I assume the moving of this is just so when we update leverage it gets the most up to date values?

raymogg (changes requested):

contracts/Account.sol
CalabashSquash: It's the opposite. At 0% full insurance pool, max leverage is 1

contracts/TracerPerpetualSwaps.sol
CalabashSquash: Yeah you and I discussed it IRL last week and agreed this was the way to go.

contracts/Account.sol
CalabashSquash: It's just a linear function where the x axis is how full the ins pool is, y axis is realMaxLeverage.

At 0% ins pool, realMaxLeverage is 1, so it passes through point (0, 1).
At INSURANCE_DELEVERAGING_CLIFF% ins pool, realMaxLeverage = baseMaxLeverage, so it passes through point (INSURANCE_DELEVERAGING_CLIFF, baseMaxLeverage).

In between, it scales linearly so I derived a generalised linear function where you can calculat the gradient by using the (x2 - x1)/(y2 - y1) formula, and b will always be 1 because that's the y intercept.

I have modified the comment to fix errors and explain this better.

raymogg (approved):

PR #35, opened 2021-04-19
Hardhat integration

raymogg:
# Motivation
Hardhat has a lot of nice features we want to be able to use. This PR moves most of our functionality over to hardhat, however does not include as in depth migration setup as previously existed. This will be added in another PR.

Also note that tests aren't utilising hardhat to the best of its abilities as we still use our custom testing library. By moving this library to hardhat, we should see reductions in our testing times.

# Changes
- remove truffle and typescript from tests
- add in hardhat
- add scripts and deploy directories


sporejack (approved):
LGTM!

CalabashSquash (approved):
:djibouti:

PR #38, opened 2021-04-22
Calculate max leverage

CalabashSquash:
### Motivation

Deleveraging is a desired feature for when insurance pool starts draining.

#### TracerPerpetualSwaps.sol
- Added `int256 INSURANCE_DELEVERAGING_CLIFF`: the upper threshold for the insurance pool when deleveraging begins
    - This is a market parameter.
- Moved settle to be above account updates in `takeOrder` and `matchOrders`.

#### Account.sol
- Added function `realMaxLeverage,` which adjusts the max leverage as insurance pool slides from TracerPerpetualSwaps.INSURANCE_DELEVERAGING_CLIFF% of target to 0% of target.


raymogg:
Extremely outdated

PR #39, opened 2021-04-23
Trader Gas Accounting

raymogg:
Copy of PR #28 migrated over to the new hardhat implementation.


sporejack:
@raymogg Good candidate for rejection now I'd say.

raymogg:
> @raymogg Good candidate for rejection now I'd say.

Yeah we haven't focused on merging this in at all yet again. Should we patch these changes on top of current state? We never fully fleshed out how we wanted gas accounting to work (if we wanted to require users to deposit or not).

raymogg:
To be resumed at a later date

PR #40, opened 2021-04-23
CI Hotfix

raymogg:
# Motivation
While it appeared CI with hardhat was doing its job, the tests actually weren't being run properly. There was a bug in the tests that meant while it appeared CI was passing, the tests wouldn't run.

# Changes
- move hardhat from dependency to dev-dependency
- fix tests


CalabashSquash (approved):
:dango:

PR #41, opened 2021-04-29
Updated to OZ4 and Solidity 0.8

raymogg:
# Motivation
Solidity 0.8 allows us to get rid of safe math and simplify a lot of our arithmetic. This will help as we move to a library based approach over the current setup.

# Changes
- bump OpenZeppelin to version 4
- bump Solidity to 0.8.0
- change maths to no longer use SafeMath and SafeERC20


CalabashSquash (approved):
:vanuatu:

sporejack (approved):
A thing of beauty!

PR #45, opened 2021-04-30
Move account state

raymogg:
# Motivation
The account contract and tracer contract are being merged to allow Tracer.sol to handle all things account state and trade execution.

# Changes
- remove account.sol
- migrate all logic and state variables into Tracer
- refactors Tracer completely
- removes redundant variables from most function calls.
- Fixes some small compilation issues with Liquidations

# Breaking changes
- Anything using account to deposit, withdraw and alter state is now broken.


sporejack (approved):
LGTM!

CalabashSquash (approved):

PR #46, opened 2021-04-30
Clean up tracer

raymogg:
# Motivation
As per #45 all account state has been migrated into the Tracer contract. This introduces opportunities to reduce the size of the codebase by removing redundant function calls, optimising pieces of code and just in general cleaning things up.

# Changes
- rename funding rate sensitivity as it isn't a constant
- remove user permissioning from the tracer (to become whitelisting of trading interfaces eventually)
- add fee rate as a constructor argument
- add todos for library functions
- general optimisations
- remove redundant use of the `market` parameter since the address of the tracer is now known in all contexts
- simplified state management on trade execution

### DIFF
--- a/contracts/TracerPerpetualSwaps.sol
+++ b/contracts/TracerPerpetualSwaps.sol
@@ -21,7 +21,7 @@ contract TracerPerpetualSwaps is
 	using LibMath for uint256;
 	using LibMath for int256;
 
-	uint256 public override FUNDING_RATE_SENSITIVITY;
+	uint256 public override fundingRateSensitivity;
 	uint256 public constant override LIQUIDATION_GAS_COST = 63516;
 	uint256 public immutable override priceMultiplier;
 	address public immutable override tracerBaseToken;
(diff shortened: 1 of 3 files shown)


sporejack (approved):
LGTM

contracts/TracerPerpetualSwaps.sol
```diff
@@ -79,7 +76,8 @@ contract TracerPerpetualSwaps is
 		address _pricingContract,
 		address _liquidationContract,
 		int256 _maxLeverage,
-		uint256 fundingRateSensitivity
+		uint256 _fundingRateSensitivity,
+		uint256 _feeRate
 	) public Ownable() {
 		pricingContract = IPricing(_pricingContract);
```
CalabashSquash, line 82: Should DeployerV1 be updated to add this param?

contracts/TracerPerpetualSwaps.sol
```diff
@@ -197,115 +194,71 @@ contract TracerPerpetualSwaps is
...
-		settle(order1User);
-		settle(order2User);
+		// settle accounts
+		settle(order1.maker);
+		settle(order2.maker);
 
 		// Update internal trade state
-		updateInternalRecords(orderPrice);
+		// note: price has already been validated here, so order 1 price can be used
+		updateInternalRecords(order1.price);
```
CalabashSquash: Should `settle` be called before `_updateAccountLeverage`?

CalabashSquash (changes requested):

CalabashSquash (approved):

PR #50, opened 2021-05-03
Move logic from Tracer to Pricing

raymogg:
# Motivation
As per the new design, the Tracer perp contract should only hold logic for actual updating of account state and trade execution. Currently it holds a lot of timing and pricing state, where it simply communicates between itself and pricing.

# Changes
- remove any timing state from Tracer and add to Pricing
- remove all cross communication except for the tracer notifying pricing of trade execution.
- removes the oracle from the tracer as it is not needed
- adds a constructor argument for oracle decimals to still support price multiplier (however this will probably be removed at some point too)
- formatting fixes

### DIFF
--- a/contracts/DeployerV1.sol
+++ b/contracts/DeployerV1.sol
@@ -15,7 +15,6 @@ contract DeployerV1 is IDeployer {
         (
             bytes32 _tracerId,
             address _tracerBaseToken,
-            address _oracle,
             address _gasPriceOracle,
             address _pricingContract,
             address _liquidationContract,
@@ -28,7 +27,6 @@ contract DeployerV1 is IDeployer {
             address,
             address,
             address,
-            address,
             int256,
             uint256,
             uint256
@@ -36,7 +34,6 @@ contract DeployerV1 is IDeployer {
         TracerPerpetualSwaps tracer = new TracerPerpetualSwaps(
             _tracerId,
             _tracerBaseToken,
-            _oracle,
             _gasPriceOracle,
             _pricingContract,
             _liquidationContract,
--- a/contracts/DeployerV1.sol
+++ b/contracts/DeployerV1.sol
@@ -20,7 +20,8 @@ contract DeployerV1 is IDeployer {
             address _liquidationContract,
             int256 _maxLeverage,
             uint256 _fundingRateSensitivity,
-            uint256 _feeRate
+            uint256 _feeRate,
+            uint256 _oracleDecimals
         ) = abi.decode(_data, (
             bytes32,
             address,
@@ -28,6 +29,7 @@ contract DeployerV1 is IDeployer {
             address,
             address,
             int256,
+            uint256,
             uint256,
             uint256
         ));
@@ -39,7 +41,8 @@ contract DeployerV1 is IDeployer {
             _liquidationContract,
             _maxLeverage,
             _fundingRateSensitivity,
-            _feeRate
+            _feeRate,
+            _oracleDecimals
         );
         tracer.transferOwnership(msg.sender);
         return address(tracer);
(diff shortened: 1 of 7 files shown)


contracts/DeployerV1.sol
```diff
@@ -15,34 +15,34 @@ contract DeployerV1 is IDeployer {
...
             address _pricingContract,
             address _liquidationContract,
             int256 _maxLeverage,
             uint256 _fundingRateSensitivity,
-            uint256 _feeRate
+            uint256 _feeRate,
+            uint256 _oracleDecimals
         ) = abi.decode(_data, (
             bytes32,
             address,
```
CalabashSquash, line 27: Should there be a setter for `_oracleDecimals` in TracerPerpSwaps.sol?

contracts/Pricing.sol
```diff
@@ -5,37 +5,110 @@ import {Types} from "./Interfaces/Types.sol";
...
+     *         a set price
+     * @param tradePrice the price the trade executed at
+     * @param tradeVolume the volume of the order
+     */
+    function recordTrade(
+        int256 tradePrice,
+        uint256 tradeVolume
+    ) external override onlyTracer {
+        int256 currentOraclePrice = oracle.latestAnswer();
+        // todo the pricing should need the oracle, not the tracer market?
```
CalabashSquash: What does this mean?

contracts/Pricing.sol
```diff
@@ -49,23 +122,32 @@ contract Pricing is IPricing {
...
         } else {
             // If an update is needed, add the market price to a running total and increment number of trades
-            price.hourlyTracerPrices[currentHour].totalPrice = price.hourlyTracerPrices[currentHour].totalPrice + marketPrice;
-            price.hourlyTracerPrices[currentHour].numTrades = price.hourlyTracerPrices[currentHour].numTrades + 1;
+            price.hourlyTracerPrices[currentHour].totalPrice =
+                price.hourlyTracerPrices[currentHour].totalPrice +
+                marketPrice;
+            price.hourlyTracerPrices[currentHour].numTrades =
+                price.hourlyTracerPrices[currentHour].numTrades +
+                1;
```
CalabashSquash, line 142: Nitpick: It would look nicer if line 140 and line 143 were on their above lines

contracts/Pricing.sol
```diff
@@ -49,23 +122,32 @@ contract Pricing is IPricing {
...
-            price.hourlyOraclePrices[currentHour].totalPrice = price.hourlyOraclePrices[currentHour].totalPrice + oraclePrice;
-            price.hourlyOraclePrices[currentHour].numTrades = price.hourlyOraclePrices[currentHour].numTrades + 1;
+            price.hourlyOraclePrices[currentHour].totalPrice =
+                price.hourlyOraclePrices[currentHour].totalPrice +
+                oraclePrice;
+            price.hourlyOraclePrices[currentHour].numTrades =
+                price.hourlyOraclePrices[currentHour].numTrades +
+                1;
         }
     }
```
CalabashSquash, line 151: Same as other comment on line 140

contracts/Pricing.sol
CalabashSquash: This is a question for the linter and whether we should configure it to do this differently.

CalabashSquash (changes requested):

contracts/DeployerV1.sol
raymogg: If the oracleDecimals changes, the priceMultiplier will change meaning all old pricing and account state is broken. May be worth enforcing this as an invariant when setting new oracles? Eg on setOracle, ensure `oracleDecimalsOld = oracleDecimalsNew`

contracts/Pricing.sol
raymogg: will remove. Hanging comment

contracts/Pricing.sol
raymogg: yep will fix the linter

PR #51, opened 2021-05-03
add whitelisting to the tracer perp

raymogg:
# Motivation
To support generalised trading interfaces, some form of whitelisting is needed per market. 

# Changes
- adds a trading whitelist mapping
- adds setter function
- requires the contract to be whitelisted to call the `matchOrders` function.


CalabashSquash (approved):
:st_kitts_nevis:

PR #52, opened 2021-05-04
Chainlink oracle adapter

raymogg:
# Motivation
In order to keep consistency across markets, an oracle adapter has been added that converts all prices reported by the oracle to a WAD format. This also means that the oracle feed can be changed without stopping and migrating the market, as an oracle using a new decimal format can be used, and this adapter will still return the WAD formatted price

# Changes
- move oracles to their own folder
- create `ChainlinkOracleAdapter.sol`
- remove redundant function from oracles
- fix the gas oracle `isStale` function to actually return the results from the upstream oracle

### DIFF
--- a/contracts/Interfaces/IOracle.sol
+++ b/contracts/Interfaces/IOracle.sol
@@ -8,7 +8,5 @@ interface IOracle {
     function isStale() external view returns (bool);
 
     function decimals() external view returns (uint8);
-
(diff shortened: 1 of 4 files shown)


sporejack (approved):
LGTM!

CalabashSquash (approved):
I like this solution to the problem :st_helena: :st_vincent_grenadines:

PR #53, opened 2021-05-04
Wrap trader struct

CalabashSquash:
### Motivation
The original motivation for this change was that we want a liquidator to be able to easily submit a receipt claim by providing the orders they made, regardless of the trading interface they were made with.

This wasn't possible if the Trader.sol contract was used, because of the way the contract's contract-specific `LimitOrder` was used. This meant that the liquidator would have to provide different details depending on whether or not they used the Trader.sol in selling their liquidation positions.

I originally added two arrays to `Types.Order`, `fillers` and `fillAmounts`, but this isn't really necessary because you can still verify the liquidator actually made trades without this (by just looking at the order(s) they made).

### Changes

**ITrader.sol**
- Added this interface for the Trader.sol contract
**Types.sol**
- Removed `LimitOrder`
- Add `nonce` to `SignedLimitOrder`
- Made `SignedLimitOrder` contain `Order`, instead of `LimitOrder`
- Added `fillers` and `fillAmounts` to `Types.Order`
    - This allows you to keep track of who has filled which amounts
**Trader.sol**
- Made changes to reflect the adding of `ITrader.sol`
- Verify targetedTracer matches actual tracer
- Replaced all `Types.LimitOrder`s with `Types.Order` and made all relevant changes

**TracerPerpetualSwaps.sol**
- Removed outdated comments
- `tradingWhitelist` is now an interface variable

**Liquidation.sol**
- **Some random changes, but kept them commented out, since I'll be doing the relevant escrow/receipt changes in a separate PR.**


contracts/Trader.sol
```diff
@@ -2,29 +2,30 @@
...
+    string private constant EIP712_DOMAIN_NAME = "Tracer Protocol";
+    string private constant EIP712_DOMAIN_VERSION = "1.0";
     bytes32 private constant EIP712_DOMAIN_SEPERATOR =
-        keccak256('EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)');
+        keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)");
 
     // EIP712 Types
-    bytes32 private constant LIMIT_ORDER_TYPE =
+    bytes32 private constant ORDER_TYPE =
         keccak256(
```
raymogg, line 23: This should also contain the `address targetTracer` field

contracts/Trader.sol
```diff
@@ -117,26 +121,25 @@ contract Trader {
...
-                block.timestamp //creation
-            );
+            Types.Order storage newOrder = orders[orderHash];
+            newOrder.maker = signedOrder.order.maker;
+            newOrder.amount = signedOrder.order.amount;
+            newOrder.price = signedOrder.order.price;
+            newOrder.filled = 0;
+            newOrder.side = signedOrder.order.side;
+            newOrder.expiration = signedOrder.order.expiration;
+            newOrder.creation = block.timestamp;
```
raymogg, line 142: also missing targetTracer setting

raymogg (changes requested):

raymogg (approved):

PR #54, opened 2021-05-04
Move Insurance state

kumar-ish:
# Motivation

The insurance contract is being made such that it is individual for each tracer -- meaning that there isn't one insurance contract that holds the insurance state for all tracers. 

# Changes

- Decouple insurance contracts from all tracers; each tracer now has an insurance contract of its own

- Change account instances to support the new TracerPerpetualSwaps contract

- Instantiate insurance contracts in deployer

- Add deposit & withdraw to perpetual swaps interface so that the insurance contract can call it

### DIFF
--- a/contracts/Insurance.sol
+++ b/contracts/Insurance.sol
@@ -51,7 +51,7 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
         collateralToken.transferFrom(msg.sender, address(this), amount);
         
         // Update pool balances and user
-        this.updatePoolAmount();
+        updatePoolAmount();
         InsurancePoolToken poolToken = InsurancePoolToken(token);
         uint256 tokensToMint;
         
@@ -80,7 +80,7 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
     function withdraw(uint256 amount) external override {
         require(amount > 0, "INS: amount <= 0");
 
-        this.updatePoolAmount();
+        updatePoolAmount();
         uint256 balance = getPoolUserBalance(msg.sender);
         require(balance >= amount, "INS: balance < amount");
 
@@ -107,7 +107,7 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
      * @notice Internally updates a given tracer's pool amount according to the tracer contract
      * @dev Withdraws from tracer, and adds amount to the pool's amount field. 
      */
-    function updatePoolAmount() external override {
+    function updatePoolAmount() public override {
         int256 base = (tracer.getBalance(address(this))).base;
         if (base > 0) {
             tracer.withdraw(uint(base));
--- a/contracts/Insurance.sol
+++ b/contracts/Insurance.sol
@@ -28,7 +28,8 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
     event InsuranceWithdraw(address indexed market, address indexed user, uint256 indexed amount);
     event InsurancePoolDeployed(address indexed market, address indexed asset);
 
-    constructor(address _tracer) {
+    constructor(address _tracer, address _perpsFactory) {
+        perpsFactory = ITracerPerpetualsFactory(_perpsFactory);
         require(perpsFactory.validTracers(_tracer), "Pool not deployed by perpsFactory");
  
         tracer = ITracerPerpetualSwaps(_tracer);
@@ -37,7 +38,7 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
         collateralAsset = tracer.tracerBaseToken();
         poolAmount = 0;
 
-        // emit InsurancePoolDeployed(_tracer, tracer.tracerBaseToken());
+        emit InsurancePoolDeployed(_tracer, tracer.tracerBaseToken());
     }
 
     /**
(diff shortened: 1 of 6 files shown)


contracts/Insurance.sol
```diff
@@ -15,122 +14,112 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
...
+    constructor(address _tracer) {
+        require(perpsFactory.validTracers(_tracer), "Pool not deployed by perpsFactory");
+ 
+        tracer = ITracerPerpetualSwaps(_tracer);
+        InsurancePoolToken _token = new InsurancePoolToken("Tracer Pool Token", "TPT");
+        token = address(_token);
+        collateralAsset = tracer.tracerBaseToken();
+        poolAmount = 0;
+
+        // emit InsurancePoolDeployed(_tracer, tracer.tracerBaseToken());
```
raymogg: May need to check with @dospore what affect this change will have on the frontend. I imagine he is watching for all events from the old insurance contract. We should probably still emit this event IMO

contracts/Insurance.sol
```diff
@@ -15,122 +14,112 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
...
      * @dev Mints amount of the pool token to the user
      * @param amount the amount of tokens to stake
-     * @param market the address of the tracer market to provide insurance
      */
-    function stake(uint256 amount, address market) external override {
-        StakePool storage pool = pools[market];
-        IERC20 token = IERC20(pool.collateralAsset);
-        require(supportedTracers[market], "INS: Tracer not supported");
+    function stake(uint256 amount) external override {
+        require(perpsFactory.validTracers(address(tracer)), "Pool not deployed by perpsFactory");
```
raymogg: Don't think this check is needed since we already validate this in the constructor

contracts/TracerPerpetualsFactory.sol
```diff
@@ -2,9 +2,9 @@
 pragma solidity ^0.8.0;
 
 import "./Interfaces/ITracerPerpetualSwaps.sol";
-import "./Interfaces/IInsurance.sol";
 import "./Interfaces/ITracerPerpetualsFactory.sol";
 import "./Interfaces/IDeployer.sol";
+import "./Insurance.sol";
```
raymogg, line 7: This will work fine for now, but just making a comment that as we flesh out our new contract dependencies, a lot of this deployment stuff will need to change. Eventually the size of the PerpFactory will get too big if we import every fully implemented contract here.

raymogg (changes requested):

contracts/Insurance.sol
```diff
@@ -15,122 +14,112 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
...
     }
 
     /**
      * @notice Allows a user to withdraw their assets from a given insurance pool
      * @dev burns amount of tokens from the pool token
      * @param amount the amount of pool tokens to burn
-     * @param market the tracer contract that the insurance pool is for.
      */
-    function withdraw(uint256 amount, address market) external override {
+    function withdraw(uint256 amount) external override {
```
CalabashSquash, line 80: Just a general thought: Should `withdraw` call `updatePoolAmount`? I realise the user can just call it themself before withdrawing, but is this extra step desired?
@raymogg

contracts/Insurance.sol
kumar-ish: perpsFactory could be changed at the time of the commit, thus requiring it -- this has been changed so that perpsFactory can't change (only set at construction) so it's already validated and the check isn't required

contracts/Insurance.sol
kumar-ish: Oversight on my part; fixed now

contracts/Insurance.sol
kumar-ish: I think it would help mitigate potential errors since it would definitely need to be there before `withdraw` is called, so I think so :+1:

contracts/Insurance.sol
raymogg: @BenjaminPatch @kumar-ish good point. It should probably also be called on deposit then hey? Otherwise ratios will be out

contracts/Insurance.sol
raymogg: @BenjaminPatch @kumar-ish good point. It should probably also be called on deposit then hey? Otherwise ratios will be out

kumar-ish: Yes, it should -- fixed now

CalabashSquash (approved):
:st_helena:

contracts/Insurance.sol
```diff
@@ -79,6 +80,7 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
     function withdraw(uint256 amount) external override {
         require(amount > 0, "INS: amount <= 0");
 
+        this.updatePoolAmount();
```
raymogg: If we are calling this internally (which we are) it makes sense to make the function public rather than external and avoid using `this.updatePoolAmount()`. This is because using the external function and calling `this.X()` makes a CALL which is more expensive.

See https://ethereum.stackexchange.com/questions/19380/external-vs-public-best-practices

raymogg (changes requested):

raymogg (approved):

PR #56, opened 2021-05-05
Base implementation of LibPerpetuals

sporejack:
# Motivation
Shared types and logic are best abstracted into their own module for various reasons. This is what `LibPerpetuals` does.

# Changes
 - Define `Order` type
 - Update the rest of the codebase to use this type instead of `Types.Order`
 - Remove `Types.Order`


contracts/Trader.sol
```diff
@@ -20,7 +21,7 @@ contract Trader is ITrader {
     // EIP712 Types
     bytes32 private constant ORDER_TYPE =
         keccak256(
-            "Order(address maker,uint256 amount,int256 price,uint256 filled,bool side,uint256 expiration,uint256 creation,address targetTracer)"
+            "Order(address maker,uint256 amount,int256 price,uint256 filled,bool side,uint256 expiration,uint256 created,address market)"
         );
```
CalabashSquash: I guess this should also change `expiration` to `expires`

CalabashSquash (changes requested):
One small change

contracts/Trader.sol
sporejack: Consult 55421b0.

contracts/Trader.sol
```diff
@@ -70,31 +78,34 @@ contract Trader is ITrader {
...
             // update order state
-            makeOrder.filled = makeOrder.filled + fillAmount;
-            takeOrder.filled = takeOrder.filled + fillAmount;
+            filled[Perpetuals.orderId(makeOrder)] = makeOrderFilled + fillAmount;
+            filled[Perpetuals.orderId(takeOrder)] = takeOrderFilled + fillAmount;
 
             // increment nonce if filled
-            bool completeMaker = makeOrder.filled == makeOrder.amount;
-            bool completeTaker = takeOrder.filled == takeOrder.amount;
+            bool completeMaker = makeOrderFilled == makeOrder.amount;
```
raymogg: These comparisons are comparing the previous filled amount (before this trade was executed).

They should compare `filled[id] == makeOrder.amount`

contracts/Trader.sol
```diff
@@ -150,7 +160,7 @@ contract Trader is ITrader {
      * @param order the limit order being hashed
      * @return an EIP712 compliant hash (with headers) of the limit order
      */
-    function hashOrder(Types.Order memory order) public view override returns (bytes32) {
+    function hashOrder(Perpetuals.Order memory order) public view override returns (bytes32) {
```
raymogg, line 163: Just raising that this function is pretty much the same as `Perpetual.orderId(order)`, however it uses `abi.encodePacked`. I know this is used for comparing EIP712 signatures so may not be worth using the `orderId` function, but leaving a comment to just raise the point.

I realise this is used by the EIP712 spec for signing so maybe it shouldn't and this should be kept as is here, either in case we change the ID implementation or if we don't want to use `abi.encodePacked` for the order id just worth raising.

raymogg (changes requested):

contracts/Trader.sol
sporejack: Does EIP-712 mandate packed encoding?

raymogg: @sporejack due to how the ethSignTypedData RPC call works, I am fairly certain it needs data to be encoded in packed form, so yeah it does.

Thats the main reason I would be pro keeping this function. Just wanted to raise it so we were all aware

sporejack: @raymogg If it *is* mandated then yes, I'd agree in that we do need this function to remain.

contracts/Trader.sol
sporejack: Consult 2c0820d.

raymogg (approved):

PR #58, opened 2021-05-05
Claim receipts

CalabashSquash:
### Motivation
`claimReceipts` was broken due to major refactoring.

### Changes
**Liquidation.sol**
- There used to be `Account.claimReceipts` and `Receipt.claimReceipts`.
    - `Receipt.claimReceipts` just calculated how much to return to the claimant, so it changed to `calcAmountToReturn`
- Make sure claimant is claiming against a whitelisted trading interface
- In `calcUnitsSold`:
    - Check all invalid conditions, and `emit InvalidClaimOrder` before continuing (instead of reverting)
    - The maths can be simplified now, especially with the new Order flow
        - all you need is two lines updating the units sold and price variables
- In `claimReceipt`:
    - Since this doesn't exist in Account.sol anymore, we need to calculate the data, and push this info to `TracerPerpetualSwaps.sol` to update account balances.
- Removed reference to market, since Liquidation acts on a per-market basis
- Once `calcUnitsSold` has finished, you can now have less than `receipt.liquidatedAmount` as the `unitsSold`, because invalid orders are ignored.


contracts/Insurance.sol
```diff
@@ -126,7 +126,10 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
      * @param market The Tracer market whose insurance pool will be drained
      * @param amount The desired amount to take from the insurance pool
      */
-    function drainPool(address market, uint256 amount) external override onlyAccount() {
+    function drainPool(
```
raymogg: I'm assuming `onlyAccount` should be changed to something to do with liquidations? `onlyLiquidation`?

contracts/Liquidation.sol
```diff
@@ -321,13 +312,12 @@ contract Liquidation is ILiquidation, Ownable {
...
             liquidatedBalance.lastUpdatedGasPrice,
-            account,
-            market
+            account
         );
         
         // Limits the gas use when liquidating 
         require(
-            tx.gasprice <= uint256(IOracle(ITracerPerpetualSwaps(market).gasPriceOracle()).latestAnswer().abs()),
+            tx.gasprice <= uint256(IOracle(tracer.gasPriceOracle()).latestAnswer().abs()),
```
raymogg, line 320: This could be changed in a future PR, but I am pretty sure the tracer market does this check too. We should move the gasPriceOracle across to the Liquidations contract and remove it from Tracer

raymogg (changes requested):

contracts/Insurance.sol
CalabashSquash: Yep good catch @raymogg

contracts/Liquidation.sol
CalabashSquash: Made ticket

PR #60, opened 2021-05-06
getLiquidationReceipt now returns a struct

CalabashSquash:
### Motivation
Instead of returning every single variable from a struct, we can just return the struct.

### Changes
- `Liquidation.getLiquidationReceipt` now returns a struct


raymogg (approved):
🥇

PR #61, opened 2021-05-06
Moved old tests to old_tests/, created new test directory

CalabashSquash:
### Motivation
Just so we can add tests now as we go.
### Changes
- Remove all of the old tests
- Added a single test file in the new `test/unit/` directory. This can be used as a reference (but probably not needed)
    - It's a basic test for `LibLiquidation.calcEscrowLiquidationAmount`
- Remove truffle and web3.js from package.json


raymogg (changes requested):
We may as well just delete the old tests? They will exist in history for when we want to reference them for writing new ones. Doesn't make sense IMO to keep them in version control though if they are being killed.

raymogg (approved):

PR #62, opened 2021-05-06
Convert Price and other params to uint256

raymogg:
# Motivation
It was decided to reduce scope and to be able to enforce more tight invariants, all pricing data would be represented as uints. This PR makes these changes, but also many of the off shoot changes that are required to be made.

# Changes
- convert all representations of price to uints
- update oracle interfaces to conform
- update maxLeverage to be uint
- update leveragedValue representations to be uint
- updated lastUpdatedGasPrice to be uint
- updated minMargin to be uint
- change all function params to match new type system. Alter some small sections of maths to be more optimised for uints
- made not of potential overflow and math checks that need to be performed
- updated liquidations and LibLiquidations to support new types. Remove unneccesary use of int256 throughout these two files and replaced with uints where possible


contracts/lib/LibBalances.sol
```diff
@@ -116,25 +99,27 @@ library Balances {
      * @param liquidationGasCost The cost to perform a liquidation
      * @param maxLeverage The maximum ratio of notional value/margin
      */
+     // todo can min margin be negative?
```
CalabashSquash: It can't

CalabashSquash (approved):
:western_sahara:

sporejack (approved):
LGTM!

PR #63, opened 2021-05-06
Base implementation of LibBalances

sporejack:
TBA


contracts/TracerPerpetualSwaps.sol
```diff
@@ -220,12 +220,11 @@ contract TracerPerpetualSwaps is
...
 		uint256 originalLeverage = userBalance.totalLeveragedValue;
+        Balances.Position memory pos = Balances.Position(userBalance.base, userBalance.quote);
 		uint256 newLeverage =
-			Balances.newCalcLeveragedNotionalValue(
-				userBalance.quote,
-				pricingContract.fairPrice(),
-				userBalance.base,
-				priceMultiplier
+			Balances.leveragedNotionalValue(
+                pos,
```
BenjaminPatch, line 226: nitpick: indentation

contracts/TracerPerpetualSwaps.sol
```diff
@@ -220,12 +220,11 @@ contract TracerPerpetualSwaps is
...
+				pricingContract.fairPrice()
```
CalabashSquash, line 236: nitpick: indentation

contracts/lib/LibBalances.sol
```diff
@@ -1,161 +1,108 @@
...
+        uint256 amount;
+        Perpetuals.Side side;
     }
 
+    function netValue(
+        Position calldata position,
+        uint256 price
+    ) public pure returns (uint256) {
+        /* cast is safe due to semantics of `abs` */
+        return uint256(position.quote.abs()) * price;
```
CalabashSquash, line 26: This should be `base * price` now

contracts/lib/LibBalances.sol
```diff
@@ -1,161 +1,108 @@
...
+
+        int256 newBase = 0;
+        int256 newQuote = 0;
+
+        if (trade.side == Perpetuals.Side.Long) {
+            newQuote = position.quote + signedAmount;
+            newBase = position.base - baseChange + fee;
+        } else if (trade.side == Perpetuals.Side.Short) {
+            newQuote = position.quote - signedAmount;
+            newBase = position.base + baseChange - fee;
```
CalabashSquash, line 100: With the new base/quote, on a long trade, base goes up, and quote goes down

CalabashSquash (changes requested):

contracts/TracerPerpetualSwaps.sol
```diff
@@ -220,12 +220,11 @@ contract TracerPerpetualSwaps is
 	function _updateAccountLeverage(address account) internal {
 		Types.AccountBalance memory userBalance = balances[account];
 		uint256 originalLeverage = userBalance.totalLeveragedValue;
+        Balances.Position memory pos = Balances.Position(userBalance.base, userBalance.quote);
```
raymogg, line 232: Happy with this for now, but what we can do eventually is probably change the balances mapping. to return a `Balances.Position` struct rather than raw base and quote. That way we don't need this in the future

raymogg (approved):

PR #64, opened 2021-05-06
Liquidation balance changes test

CalabashSquash:
### Motivation
To test `LibLiquidation.liquidationBalanceChanges`

### Changes
- Re-did the `hardhat-deploy` scripts in `deploy/`.
- Started using `hardhat-deploy` plugin
- Improved the tests for `calcLiquidationEscrowAmount`
- Wrote tests for `liquidationBalanceChanges`


contracts/lib/LibLiquidation.sol
```diff
@@ -97,7 +97,7 @@ library LibLiquidation {
         int256 maxSlippage,
         int256 avgPrice,
         LibLiquidation.LiquidationReceipt memory receipt
-    ) internal pure returns (uint256) {
+    ) public pure returns (uint256) {
```
raymogg, line 100: Are these only public purely for testing purposes? Wondering if there is an easier way to test libs than jsut having to make the functions public

raymogg (approved):

contracts/lib/LibLiquidation.sol
CalabashSquash: I couldn't find any way. We were talking about it in slack yesterday and figured it was fine. Also there's no risk of them being public since they are pure

PR #65, opened 2021-05-06
Collateral library

raymogg:
# Motivation
In order to support WAD maths, rather than each market having some multiply factor, it is much easier to convert each token to a WAD format on deposit (so all internal state is WAD) and then convert it back to a raw token format on withdraw.

# Changes
- introduces base token decimals, since ERC20 tokens are not required to implement the `decimals()` function
- convert token amount to WAD amount on deposit and use this for account state
- convert WAD amount to token amount on withdraw.

# Further Changes
In order for this to be fully supported, some new maths will be needed on the conversion from WAD to raw token amount on withdraw. This is because the correct amount of rounding needs to be done to prevent rounding issues causing withdraw issues. WAD to raw token amount conversions should round down to leave dust in the contract itself.

### DIFF
--- a/contracts/Interfaces/ITracerPerpetualSwaps.sol
+++ b/contracts/Interfaces/ITracerPerpetualSwaps.sol
@@ -27,6 +27,8 @@ interface ITracerPerpetualSwaps {
 
     function tracerBaseToken() external view returns (address);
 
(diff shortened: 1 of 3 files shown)


CalabashSquash (approved):
:mali:

sporejack (approved):
LGTM!

PR #70, opened 2021-05-08
Deployment of Pricing per market

CalabashSquash:
### Motivation
There were some parts of a market deployment that weren't yet being done

### Changes
 - Removed insurance as being stored in factory, since it will change per market
 - Removed Pricing as a parameter to `TracerPerpetualSwaps` constructor, because we will deploy one new Pricing contract per market, which takes market address as a param
     - So now we have to set Pricing contract after deploying


raymogg (approved):
LGTM, although we should probably change the `deploy/FullDeploy` file to be something else. Not sure how we want to run the deploy structure with hardhat (multiple files, one per conntract with a full deploy file?)

contracts/TracerPerpetualSwaps.sol
```diff
@@ -71,16 +70,13 @@ contract TracerPerpetualSwaps is
...
 		address _liquidationContract,
 		uint256 _maxLeverage,
 		uint256 _fundingRateSensitivity,
 		uint256 _feeRate,
 		uint256 _oracleDecimals
 	) Ownable() {
-		pricingContract = IPricing(_pricingContract);
-		// dont convert to interface as we don't need to interact
-		// with the contract
+		// dont convert to interface as we don't need to interact with the contract
```
sporejack:
```suggestion
		// don't convert to interface as we don't need to interact with the contract
```

sporejack (changes requested):
One nitpick, otherwise this should be good to go.

PR #73, opened 2021-05-10
Add new account type to `LibBalances`

sporejack:
# Motivation
Previously, account balance state relied on two separate fields for both base and quote. With the new `Position` type, this is now incorrect. Additionally, the `Types.AccountBalance` type can simply be called `Account` now.

# Changes
 - Define `Account` type in `LibBalances`
 - Remove `Types.AccountBalance`
 - Fix codebase to use new changes

### DIFF
--- a/contracts/Liquidation.sol
+++ b/contracts/Liquidation.sol
@@ -160,16 +160,11 @@ contract Liquidation is ILiquidation, Ownable {
         return amountToReturn;
     }
 
-
     /**
      * @notice Allows a trader to claim escrowed funds after the escrow period has expired
      * @param receiptId The ID number of the insurance receipt from which funds are being claimed from
      */
-    function claimEscrow(uint256 receiptId)
-        public
-        override
-        onlyTracer
-    {
+    function claimEscrow(uint256 receiptId) public override onlyTracer {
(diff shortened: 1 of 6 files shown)


CalabashSquash (approved):
:niue:

contracts/Liquidation.sol
```diff
@@ -334,14 +335,13 @@ contract Liquidation is ILiquidation, Ownable {
...
 
         /* Liquidated account's balance */
-        Types.AccountBalance memory liquidatedBalance =
-            tracer.getBalance(account);
+        Balances.Account memory liquidatedBalance = tracer.getBalance(account);
 
         uint256 amountToEscrow =
             verifyAndSubmitLiquidation(
-                liquidatedBalance.quote,
+                liquidatedBalance.position.quote,
```
raymogg, line 342: Are you planning to modify all these calls that separate out base and quote into arguments so that they simply take in a position? Would be worthwhile in a new PR imo

raymogg (approved):

contracts/Liquidation.sol
sporejack: @raymogg Yes to both statements.

PR #75, opened 2021-05-10
add wad maths to liquidation library

raymogg:
# Motivation
The liquidations library still had a `PERCENT_PRECISION` factor that was being used for maths. This needed to be replaced by WAD maths

# Changes
- remove `PERCENT_PRECISION`
- expect all params to be in WAD value on input
- interpret `maxSlippage` and `percentSlippage` as whole number percent representations


contracts/lib/LibLiquidation.sol
```diff
@@ -3,13 +3,14 @@ pragma solidity ^0.8.0;
 
 import "./LibMath.sol";
 import "prb-math/contracts/PRBMathUD60x18.sol";
+import "prb-math/contracts/PRBMathSD59x18.sol";
+import "hardhat/console.sol";
```
sporejack:
```suggestion

```

contracts/lib/LibLiquidation.sol
```diff
@@ -129,29 +135,25 @@ library LibLiquidation {
...
-                    amountExpectedFor;
+                amountToReturn = amountSoldFor - amountExpectedFor;
+            }
+            if (amountToReturn <= 0) {
+                return 0;
             }
+            // multiply by 100 as we expect this to be a percent as an integer eg 50% = 50
+            percentSlippage =
+                PRBMathUD60x18.div(amountToReturn, amountExpectedFor) *
+                100;
```
sporejack, line 151: Is this valid? Why doesn't `100` need to be WAD'd also?

sporejack (changes requested):
1 nitpick, 1 question.

contracts/lib/LibLiquidation.sol
raymogg: Easiest way to think about this is in regular integer math. `amountToReturn / amountExpectedFor` returns a percentage as a decimal representation (eg `1/2 = 0.5` however we provide max slippage as an integer representation of the max slippage (eg 50% max slipapge). So to make sure the check for `percentSlippage > maxSlippage` is valid, we do this multiplication. Note we could also just make maxSlippage a decimal representation if we wanted.

We then divide the amount to return by 100 as when you calculate an actual percent value, you want to use the decimal representation (eg 50% of 5 = 0.5 * 5 = 2.5)

sporejack: Did we want to comment this somewhere or leave it for now?

CalabashSquash (approved):

contracts/lib/LibLiquidation.sol
raymogg: @sporejack I'll record this in our documentation

PR #81, opened 2021-05-11
Base Implementation of `LibPrices`

sporejack:
# Motivation
Cleaner pricing logic and types (the most important of all!).

# Changes
 - Add `FundingRateInstant` type
 - Remove `Types.FundingRate` type
 - Fix contracts to use new funding rate type


contracts/Pricing.sol
```diff
@@ -268,43 +257,34 @@ contract Pricing is IPricing {
...
+        returns (Prices.FundingRateInstant memory)
     {
-        Types.FundingRate memory fundingRate = fundingRates[index];
-        return (
-            fundingRate.recordTime,
-            fundingRate.recordPrice,
-            fundingRate.fundingRate,
-            fundingRate.fundingRateValue
-        );
+        Prices.FundingRateInstant memory fundingRate = fundingRates[index];
```
raymogg: Why create a new FundingRateInstant here? Should just return the fundingRate param in memory

contracts/Pricing.sol
```diff
@@ -268,43 +257,34 @@ contract Pricing is IPricing {
...
     {
-        Types.FundingRate memory fundingRate = insuranceFundingRates[index];
-        return (
-            fundingRate.recordTime,
-            fundingRate.recordPrice,
-            fundingRate.fundingRate,
-            fundingRate.fundingRateValue
-        );
+        Prices.FundingRateInstant memory fundingRate =
+            insuranceFundingRates[index];
```
raymogg: Same as above. No need to create a whole new FundingRateInstant struct

raymogg (changes requested):

contracts/Pricing.sol
sporejack: Consult e317c30.

contracts/Interfaces/Types.sol
```diff
@@ -2,23 +2,18 @@
...
 
-    struct HourlyPrices {
-        uint256 totalPrice;
-        uint256 numTrades;
-    }
-
-    struct PricingMetrics {
-        Types.HourlyPrices[24] hourlyTracerPrices;
-        Types.HourlyPrices[24] hourlyOraclePrices;
+    struct PricesMetrics {
```
CalabashSquash: Nitpick: Why is `PriceMetrics` in here and not in `LibPricing.sol`?

CalabashSquash: Also, is `struct PriceMetrics` even being used anymore?

sporejack: Shouldn't even exist; good catch.

sporejack: Consult cb37d4f

contracts/TracerPerpetualSwaps.sol
```diff
@@ -359,7 +332,7 @@ contract TracerPerpetualSwaps is
      * @dev Ensures the account remains in a valid margin position. Will throw if account is under margin
      *      and the account must then be liquidated.
      * @param account the address to settle.
-     * @dev This function aggregates data to feed into account.sol"s settle function which sets
+     * @dev This function aggregates data to feed into account.sols settle function which sets
```
raymogg, line 335: Comment needs changing. not part of this PR but just noting

contracts/lib/LibMath.sol
```diff
@@ -12,4 +12,21 @@ library LibMath {
     function abs(int256 x) internal pure returns (int256) {
         return x > 0 ? int256(x) : int256(-1 * x);
     }
+
```
raymogg, line 15: I am fairly certain the PRB math library has quite efficient implementation of these. Also since fundingRate will most likely end up being a WAD value (since the price data its based off is WAD values) we will want to eventually use that

raymogg: actually re thinking, the fact that they are WAD values shouldn't matter. Its more a question of which implementation is more efficient

contracts/lib/LibPrices.sol
```diff
@@ -0,0 +1,121 @@
...
+                instantUnderlying += currTimeWeight;
+                cumulativeUnderlying += currTimeWeight * currUnderlyingPrice;
+            }
+
+            if (instantDerivative == 0) {
+                return TWAP(0, 0);
+            } else {
+                return
+                    TWAP(
+                        cumulativeUnderlying / instantUnderlying,
```
raymogg: These should be WAD divides since they are simply a scaled underlying average price, which is a WAD value. 

If we don't use WAD division, the returned value will not be a WAD and will throw out a lot.

contracts/lib/LibMath.sol
sporejack: @raymogg Honestly all for a non-homemade solution for these -- I'd prefer to not have to write them myself. I agree about the correctness not being impacted and that we should then prioritise efficiency.

contracts/Pricing.sol
```diff
@@ -217,46 +214,38 @@ contract Pricing is IPricing {
...
+    function setFundingRate(int256 fundingRate, int256 cumulativeFundingRate)
+        internal
+    {
+        fundingRates[currentFundingIndex] = Prices.FundingRateInstant(
             block.timestamp,
-            marketPrice,
             fundingRate,
-            fundingRateValue
+            cumulativeFundingRate
         );
```
CalabashSquash, line 233: So just to clarify, is the reason for removing the `marketPrice` parameter just because its not really necessary when you'r ekeeping track of funding rates at different points in time?

contracts/lib/LibPrices.sol
```diff
@@ -0,0 +1,121 @@
...
+        uint256 newLeverage
+    ) public pure returns (uint256) {
+        uint256 delta = oldLeverage - newLeverage;
+
+        if (oldLeverage < newLeverage) {
+            return globalLeverage + delta;
+        } else {
+            if (delta == globalLeverage) {
+                return 0;
+            } else {
```
CalabashSquash: I'm not sure I understand this. If the change == the global leveraged value, the global leveraged value becomes 0?

What if delta == 100, globalLeverage == 100, shouldn't globalLeverage become 200?

CalabashSquash: Consider this e.g.
newLeverage = 200
oldLeverage = 100
then delta = -100 (but delta is unsigned so underflow)

Is this a concern?

CalabashSquash (changes requested):
:seychelles:

contracts/lib/LibPrices.sol
sporejack: So, in the updated version of this function body:

```solidity
int256 delta = int256(newLeverage - oldLeverage);

if (delta >= 0) { /* leverage has increased or remained the same */
    return globalLeverage + uint256(delta);
} else { /* leverage has decreased */
    if (uint256(delta) > globalLeverage) { /* handle underflow */
        return 0;
    } else {
        return globalLeverage - uint256(delta);
    }
}
```

This implements a floor for `globalLeverage`, due to it being unsigned.

raymogg (approved):

contracts/lib/LibPrices.sol
sporejack: Consult 7d054b5.

contracts/lib/LibPrices.sol
sporejack: > Is this a concern?

Yes, I believe the newer version of this block addresses this issue. Keen to hear your input.

contracts/Pricing.sol
sporejack: `marketPrice` is simply a redundant field -- we came to the conclusion it was always redundant (even in its previous usage).

contracts/lib/LibPrices.sol
CalabashSquash: I think this would still not work, because:
- (discussed in person:) int256(newLeverage - oldLeverage) would still underflow before being case to int256
- in the `else` block, you cast `delta` to `uint256`, but `delta` has to be <0 to enter this branch, leading to an invalid cast.

raymogg (approved):

contracts/lib/LibPrices.sol
```diff
@@ -79,7 +79,7 @@ library Prices {
                 /* handle underflow */
                 return 0;
             } else {
-                return globalLeverage - uint256(delta);
+                return globalLeverage - (oldLeverage - newLeverage);
```
raymogg: Why do this? Also this changes the logic right, since `delta = newLeverage - oldLeverage`, so previously this was `globalLeverage - (newLeverage - oldLeverage)`

contracts/lib/LibPrices.sol
```diff
@@ -0,0 +1,127 @@
...
+        PriceInstant[] memory oraclePrices
+    ) public pure returns (TWAP memory) {
+        uint256 instantDerivative = 0;
+        uint256 cumulativeDerivative = 0;
+        uint256 instantUnderlying = 0;
+        uint256 cumulativeUnderlying = 0;
+
+        for (uint256 i = 0; i < 8; i++) {
+            uint256 currTimeWeight = 8 - i;
+            uint256 j = 8 - i;
```
CalabashSquash, line 92: Did we come to a conclusion for how this should be fixed (TWAP not taking into account the `hour`)?

sporejack: @BenjaminPatch can either fold it into this PR or file a separate one (given there's a separate ticket).

contracts/lib/LibPrices.sol
```diff
@@ -68,19 +68,15 @@ library Prices {
         uint256 oldLeverage,
         uint256 newLeverage
     ) public pure returns (uint256) {
-        int256 delta = int256(newLeverage) - int256(oldLeverage);
+        bool leverageHasIncreased = newLeverage > oldLeverage;
```
raymogg, line 71: can probably be simplified I think. Should only need one bool it seems, as the only edge case is where `newLeverage = oldLeverage` in which case `newLeveraged - oldLeverage = 0` anyway.

Current way is more explicit, but I think you should just be able to do 

`if (leverageHasIncreased) { add to leverage } else { decrease leverage }`

PR #83, opened 2021-05-11
factor out minting and burning math to lib

raymogg:
# Motivation
In order to be able to better unit test some of the insurance logic, an insurance lib contract has been created

# Changes
- move minting and burning maths logic to the insurance lib
- renamed `poolAmount` to be `collateralAmount` as pool amount was quite ambiguous


sporejack (approved):
LGTM!

PR #85, opened 2021-05-12
Added fullLiquidate. Added check to ensure amount < agent's base

CalabashSquash:
### Motivation
In the case of full liquidation, the liquidator would have to provide an exact amount to liquidate, in wei. This could be quite difficult, as small amounts of wei could be lost/stuck in the balance. It also means that a liquidator would risk rounding up too much if they were trying to get a complete full liquidation. Overall just bad UX in my opinion.

### Changes
- Added `fullLiquidate`
- Added check to ensure amount < agent's base

### DIFF
--- a/contracts/Interfaces/ILiquidation.sol
+++ b/contracts/Interfaces/ILiquidation.sol
@@ -22,6 +22,8 @@ interface ILiquidation {
         view
         returns (LibLiquidation.LiquidationReceipt memory);
 
+    function fullLiquidate(address account) external;
+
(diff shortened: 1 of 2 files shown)


raymogg (commented):
Happy with the changes, but wondering if this should be an application (eg keeper bot) vs protocol layer change?

A keeper or liquidator could read `tracer.getbalance(account).position.base` and provide the exact amount themselves. I doubt people are going to be liquidating by calling the contracts directly

PR #88, opened 2021-05-12
Liquidation test cases

CalabashSquash:
### Motivation
Good idea to start writing test stubs
Also bug fixes

### Changes
- `claimReceipts` -> `claimReceipt`
- Moved some of the receipt verification from `claimReceipt` into `verifyAndSubmitLiquidation`
    - This is just because I figured all verification should go into the function with "verify" in its name
    - It might even be worth creating two functions: `verifyLiquidation` and `submitLiquidation`
- Fixed bug in `claimReceipt`, where the insurance's balance wouldn't be updated by Liquidation.sol after `drainPool` was called
- Fixed some minor bugs in `LibLiquidation.liquidationBalanceChanges`
- Added test stubs for liquidation


sporejack (approved):
LGTM!

raymogg (approved):

test/functional/Liquidation.js
```diff
@@ -0,0 +1,96 @@
+const { expect } = require("chai")
+const { ethers, getNamedAccounts, deployments } = require("hardhat")
+
+describe("Liquidation functional tests", async () => {
```
sporejack, line 4: It'd be nice to nest `describe` calls, like so:

```javascript
describe("Liquidation", async () => {
    describe("someLiqFunc", async () => {
        context("When called with foobar", async () => {
            it("Passes", async () => {
                /* ... */
            });
        });
    });
});
```

sporejack (changes requested):

test/functional/Liquidation.js
CalabashSquash: I don't know if I reckon that's any better though. The output would require reading the extra "Passes" line to see what the expected result of that test case is.

test/functional/Liquidation.js
```diff
@@ -0,0 +1,96 @@
...
+    })
+
+    context("liquidate", async () => {
+        it("Reverts when liquidation would put liquidator below minimum margin", async () => {})
+
+        it("Reverts when agent isn't below margin", async () => {})
+
+        it("Reverts when gas price is above fast gas price", async () => {})
+
+        it("Reverts on negative liquidation amount", async () => {})
```
raymogg: also reverts on liquidation amount = 0

test/functional/Liquidation.js
```diff
@@ -0,0 +1,96 @@
...
+
```
CalabashSquash:
```suggestion

it("Reverts on liquidation amount == 0", async () => {})
```

test/functional/Liquidation.js
CalabashSquash: Added

PR #90, opened 2021-05-13
Moved out insurance pool draining into its own function

CalabashSquash:
### Motivation
Having a very big block of code inside a function made me think it should probably be separated into its own function

### Changes
- Moved the logic that deals with draining insurance pool during excess slippage into its own function, `drainInsurancePoolOnLiquidation`


raymogg (approved):

raymogg (approved):

sporejack (approved):
LGTM!

PR #91, opened 2021-05-13
Fix liquidation circular dependency

CalabashSquash:
### Motivation
We needed to deploy Liquidaiton.sol in factory, but adding it tipped the contract size way over limit.

So I went through and added Deployers, to reduce contract size.

### Changes
- Added `PricingDeployerV1.sol`, `LiquidationDeployerV1.sol` and `InsuranceDeployerV1.sol`
    - These just deploy their respective contracts.
- Fixed deployment so now TracerPerpetualSwaps.sol does not require liquidation on construction, but is instead set after Liquidation gets deployed.


raymogg (commented):
Nitpick but lets put all our deployers into a deployers/ folder since there is multiple of them now.

contracts/TracerPerpetualsFactory.sol
```diff
@@ -22,28 +26,43 @@ contract TracerPerpetualsFactory is Ownable, ITracerPerpetualsFactory {
 
     event TracerDeployed(bytes32 indexed marketId, address indexed market);
 
-    constructor(address _deployer, address _governance) {
+    constructor(
+        address _deployer,
```
raymogg: `perpsDeployer`?

contracts/TracerPerpetualsFactory.sol
```diff
@@ -66,13 +86,22 @@ contract TracerPerpetualsFactory is Ownable, ITracerPerpetualsFactory {
 
         // Instantiate Insurance contract for tracer
         address insurance =
-            address(new Insurance(address(market), address(this)));
-
-        address pricing = address(new Pricing(market, insurance, oracle));
+            IInsuranceDeployer(insuranceDeployer).deploy(market, address(this));
```
raymogg, line 92: Is there any worry about ownership of these contracts? Because in this flow they will be owned by the TracerPerpsFactory and not the "deployer".

CalabashSquash: Yep we should be transferring ownership of Ins, pricing, and Liq. I will add this in.

contracts/TracerPerpetualsFactory.sol
CalabashSquash: Changed

raymogg (approved):

PR #92, opened 2021-05-17
Remove reverts in Trader.sol

raymogg:
# Motivation
Our off chain order matching engine does not want transactions to revert if two transactiosn are unable to be processed as this would create a massive DoS attack vector. As such, invalid transactions should just be skipped over.

# Changes
- updates trader validation code to return bools rather than revert
- use a low level call when calling `matchOrders` on the Tracer perp rather than a direct call (which may revert)


sporejack (approved):
LGTM!

CalabashSquash (approved):

PR #93, opened 2021-05-17
Add fees

kumar-ish:
## Motivation 

We've had `feeRate` in the contracts for some time now, but have yet to implement fees in make/take orders. This PR implements that.

## Changes
- Take fee from quote (the DAI in a ETH/DAI market pair)
- Add `feeReceiver`, the address of the person able to withdraw fees; add ability to change `feeReceiver`.
- Add ability to withdraw fees


contracts/TracerPerpetualSwaps.sol
```diff
@@ -204,24 +209,44 @@ contract TracerPerpetualSwaps is
         Balances.Account storage account1 = balances[order1.maker];
         Balances.Account storage account2 = balances[order2.maker];
 
+        int256 fee =
+            PRBMathUD60x18
+                .mul(uint256(quoteChange), uint256(feeRate.toInt256()))
```
raymogg: Shouldn't need to cast `feeRate` to a int256 here

contracts/TracerPerpetualSwaps.sol
```diff
@@ -73,7 +76,8 @@ contract TracerPerpetualSwaps is
         address _gasPriceOracle,
         uint256 _maxLeverage,
         uint256 _fundingRateSensitivity,
-        uint256 _feeRate
+        uint256 _feeRate,
```
CalabashSquash, line 79: The natspec comment for this should be a bit more clear IMO.

It is vague one what exactly `feeRate` is. e.g. it doesn't specify that it is a percentage, times 10^18. e.g. `50*10^18 == 50% fee`. But based on what line 212 seems to say, it should be more like `0.5*10^18 == 50% fee`

or just some way to describe explicitly what it should be

contracts/TracerPerpetualSwaps.sol
```diff
@@ -204,24 +209,44 @@ contract TracerPerpetualSwaps is
...
+                quoteChange -
+                fee;
             account1.position.base = account1.position.base + _fillAmount;
 
             // user 2 is short. Increase quote, decrease base
-            account2.position.quote = account2.position.quote + quoteChange;
+            account2.position.quote =
+                account2.position.quote +
+                quoteChange +
+                fee;
```
raymogg, line 230: I think this should also be subtracting the fee here

contracts/TracerPerpetualSwaps.sol
```diff
@@ -204,24 +209,44 @@ contract TracerPerpetualSwaps is
...
             account2.position.base = account2.position.base - _fillAmount;
         } else {
             // user 1 is short. Increase quote, decrease base
-            account1.position.quote = account1.position.quote + quoteChange;
+            account1.position.quote =
+                account1.position.quote +
```
raymogg, line 235: same as above, should be subtracting the fee here

CalabashSquash (changes requested):

contracts/TracerPerpetualSwaps.sol
CalabashSquash: Are you sure? Since the user is going short, we would want to decrease their margin a bit.

EDIT: I mixed it up and now agree that you want to decrease quote on both cases.

contracts/TracerPerpetualSwaps.sol
```diff
@@ -204,24 +209,44 @@ contract TracerPerpetualSwaps is
...
 
+        int256 fee =
+            PRBMathUD60x18
+                .mul(uint256(quoteChange), uint256(feeRate.toInt256()))
+                .toInt256();
+
         /* TODO: handle every enum arm! */
         if (order1.side == Perpetuals.Side.Long) {
             // user 1 is long. Increase base, decrease quote
-            account1.position.quote = account1.position.quote - quoteChange;
```
sporejack, line 210: Just flagging this for review and future consideration, but this function could largely be extracted out into a pure library function that operates solely on `Balances.Position`.

contracts/TracerPerpetualSwaps.sol
```diff
@@ -510,6 +535,22 @@ contract TracerPerpetualSwaps is
...
+        feeReceiver = receiver;
+        emit FeeReceiverUpdated(receiver);
+    }
+
+    function withdrawFee() public override {
+        require(
+            feeReceiver == msg.sender,
+            "Only feeReceiver can withdraw fees"
+        );
+        fees = 0;
```
sporejack, line 550:
```suggestion
```

sporejack (changes requested):
Some comments to consider.

contracts/TracerPerpetualSwaps.sol
kumar-ish: You are right, am a silly goose

contracts/TracerPerpetualSwaps.sol
kumar-ish: Done so now

CalabashSquash: I can't see it

CalabashSquash (changes requested):

CalabashSquash (approved):
:botswana:

raymogg (approved):

sporejack (approved):
LGTM

PR #95, opened 2021-05-18
Implement unit test stubs for `LibBalances`

sporejack:
TBA


raymogg (approved):

contracts/lib/LibBalances.sol
```diff
@@ -103,6 +103,7 @@ library Balances {
 
         uint256 notionalValue = netValue(position, price);
 
+        // todo confirm that liquidation gas cost should be a WAD value
         uint256 liquidationGasCost = liquidationCost * 6;
 
         uint256 minimumMarginWithoutGasCost = notionalValue / maximumLeverage;
```
CalabashSquash: I'm wondering if this should be PRBMath.div? maximumLeverage, I believe, is a WAD value, as is notionalValue.

raymogg: @BenjaminPatch good catch I think you are right on maximumLeverage being a WAD value. Will check and update

CalabashSquash (approved):

PR #96, opened 2021-05-18
Remove nonce and add basic unit test stubs

raymogg:
# Motivation
Started out with attempting to add some working unit tests for the Trader. It was soon apparent that the nonce was not needed and lots of code could be stripped.

# Changes
- write some working tests and stub others
- remove nonce


sporejack (approved):
LGTM!

test/unit/Trader.js
```diff
@@ -0,0 +1,99 @@
...
+                )
+            })
+        })
+        context("When the maker and taker array lengths differ", async () => {
+            it("reverts", async () => {
+                await expect(trader.executeTrade([], [])).to.be.revertedWith(
+                    "TDR: Received empty arrays"
+                )
+            })
+        })
```
CalabashSquash, line 63: Just noting that while these might look like they are done, they will have to be re-written when you go to write the rest of the tests.

test/unit/Trader.js
```diff
@@ -0,0 +1,99 @@
...
+        context("When a single order signature is incorrect", async () => {
+            it("skips that order pairing", async () => {})
+        })
+        context("When an order already exists", async () => {
+            it("does not create a new order", async () => {})
+
+            it("processes the order as is", async () => {})
```
CalabashSquash, line 70: Why do we want it to "process the order as is" if an identical one already exists?

Or will the second test here make the same order with a different `created` time?

test/unit/Trader.js
```diff
@@ -0,0 +1,99 @@
...
+    })
+
+    describe("verifySignature", async () => {
+        context(
+            "When called with a valid signedOrder and signature data",
+            async () => {
+                it("returns true", async () => {})
+            }
+        )
+
```
CalabashSquash, line 94: What about a test for when `Order.created` is after current timestamp?

EDIT: Nevermind, this is tested in LibPerpetuals.

CalabashSquash (changes requested):
Just two comments that might be fine to ignore.

test/unit/Trader.js
raymogg: I guess the wording isn't 100% clear. What I want to test here is if the contract has already received an order before, it doesn't "recreate" that order. It will simply check how much of it has been filled and execute it against its paired order if possible.

PR #97, opened 2021-05-19
Unit Test LibPerpetuals

raymogg:
# Motivation
Adds unit testing for all branches in lib perpetuals.

# Changes
- adds a mock lib perpetuals contract
- adds unit tests

### DIFF
--- a/contracts/test/LibPerpetualsMock.sol
+++ b/contracts/test/LibPerpetualsMock.sol
@@ -1,0 +1,14 @@
+pragma solidity ^0.8.0;
+
+import "../lib/LibPerpetuals.sol";
+
+contract LibPerpetualsMock {
+     function canMatch(
+        Perpetuals.Order calldata a,
+        uint256 aFilled,
+        Perpetuals.Order calldata b,
+        uint256 bFilled
+    ) public view returns(bool) {
+        return Perpetuals.canMatch(a, aFilled, b, bFilled);
+    }
+}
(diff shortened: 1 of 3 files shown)


kumar-ish (approved):
:surfer:

contracts/test/LibPerpetualsMock.sol
```diff
@@ -0,0 +1,14 @@
...
+
+import "../lib/LibPerpetuals.sol";
+
+contract LibPerpetualsMock {
+    function canMatch(
+        Perpetuals.Order calldata a,
+        uint256 aFilled,
+        Perpetuals.Order calldata b,
+        uint256 bFilled
+    ) public view returns (bool) {
```
sporejack: Seeking clarification here on whether these should be declared `external` or `public`?

sporejack (approved):
Approved; left a comment for discussion.

contracts/test/LibPerpetualsMock.sol
CalabashSquash: Would say it doesn't matter since they are just used for testing

CalabashSquash (approved):
:morocco:

contracts/test/LibPerpetualsMock.sol
raymogg: @sporejack probably should be external, I'll change it just for consistency. But as Ben mentioned, this is just for testing so its not too critical.

PR #98, opened 2021-05-19
Unit test stubs for Insurance

raymogg:
# Motivation
LibInsurance is unit tested but the behaviour of Insurance itself isn't. This PR simply adds stubs for these tests.

Changes
- adds stubs for all unit tests


sporejack (approved):
LGTM!

PR #99, opened 2021-05-20
290 add fee receiver to tracer constructor

dospore:
### Motivation
- deploy script broke when we add feeReceiver to constructor

### Changes
- add feeReceiver to tracer deploy constructor
- deploy pricingLib and link it during factory deploy


sporejack (approved):
LGTM!

PR #100, opened 2021-05-20
Fix hourly tracer prices indexing

CalabashSquash:
### Motivation
On line 135 of `Pricing.sol`, it tries to access `currentHour` index of `hourlyTracerPrices`, but `hourlyTracerPrices` has not been initialised so that is an invalid index.

### Changes
- make `hourlyTracerPrices` and `hourlyOraclePrices` statically sized to be a length of 24


raymogg (approved):

sporejack (approved):
LGTM! 💯

PR #102, opened 2021-05-20
Full deployment

CalabashSquash:
- Deploy a tracer market as part of the `hardhat-deploy` deployments

**Please note**
That `deploy/GetIntoLiquidatablePosition.js` is still a WIP.

### DIFF
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -23,7 +23,7 @@ module.exports = {
         deployer: 0,
         acc1: 1,
         acc2: 2,
-        acc3: 3
+        acc3: 3,
     },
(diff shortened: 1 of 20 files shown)


hardhat.config.js
```diff
@@ -21,6 +21,9 @@ module.exports = {
     },
     namedAccounts: {
         deployer: 0,
+        acc1: 1,
```
raymogg, line 24: Just on these, idk if having names accounts is worth? You can also just pull in the accounts directly and index into the accounts array to access accounts

raymogg (approved):

PR #104, opened 2021-05-20
Divide by zero on average price

CalabashSquash:
### Motivation

Since #100 (making the hourly prices statically sized to 24 elements), getHourlyAvgTracerPrice and getHourlyAvgOraclePrice don't benefit from checking the length of hourlyTracerPrices and hourlyOraclePrices respectively.

However, they do need to check for a divide-by-zero, if number of trades in the hour is zero.

I believe this can be done simply like in this PR, but there is probably some Pricing thing that might make this more complicated.
### Changes
- Removed checking the length of hourlyTracerPrices and hourlyOraclePrices in getHourlyAvgTracerPrice and getHourlyAvgOraclePrice
- check for a divide-by-zero, if number of trades in the hour is zero.

### DIFF
--- a/contracts/Pricing.sol
+++ b/contracts/Pricing.sol
@@ -311,11 +311,7 @@ contract Pricing is IPricing, Ownable {
         override
         returns (uint256)
     {
-        if (hour >= hourlyTracerPrices.length) {
-            return 0;
-        } else {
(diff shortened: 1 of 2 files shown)


sporejack (approved):
LGTM!

PR #105, opened 2021-05-20
Smockit testing

raymogg:
# Motivation
To enable easier testing, smockit can be used to mock contracts.

# changes
- adds smockit
- uses smockit in Insurance tests
- implement insurance tests
- altered comment in Insurance
- small bug fix in `InsurancePoolToken.sol`


hardhat.config.js
```diff
@@ -14,6 +14,11 @@ module.exports = {
                 enabled: true,
                 runs: 1000,
             },
+            outputSelection: {
+                "*": {
+                    "*": ["storageLayout"],
+                },
+            },
```
CalabashSquash: what does this do?

test/unit/Insurance.js
```diff
@@ -1,101 +1,390 @@
...
+        const tracerContractFactory = await ethers.getContractFactory(
+            "TracerPerpetualSwaps",
+            {
+                libraries: {
+                    Balances: libBalances.address,
+                    Perpetuals: libPerpetuals.address,
+                    Prices: libPrices.address,
+                },
+            }
+        )
```
CalabashSquash: These should be in `deploy/FullDeploy.js` now

CalabashSquash (changes requested):
:bolivia:

hardhat.config.js
raymogg: Smockit needs access to the internal storage layout to help it mock calls, and this enables that. See: https://github.com/ethereum-optimism/optimism/tree/develop/packages/smock

test/unit/Insurance.js
raymogg: @BenjaminPatch including the mock contract and what not?

hardhat.config.js
raymogg: Also I was wrong on the above, its only needed for `smoddit` which we don't use yet. Have removed this config for now

CalabashSquash (approved):

PR #107, opened 2021-05-20
Add check for no notional value on minimumMargin

CalabashSquash:
### Motivation
If the user has no position, their minimum margin should be zero

### Changes
- Added check in `Balances.minimumMargin(...)` to return 0 if notional value == 0


sporejack (approved):
LGTM!

PR #108, opened 2021-05-21
Fix gas oracle precision units

CalabashSquash:
### Motivation
With the shift to WAD, GasOracle was completely broken

### Changes
- Replaced `latestAnswer()` logic with new logic that works with WAD.


raymogg (approved):
lgtm

PR #112, opened 2021-05-24
Wadify Funding Sensitivity

raymogg:
# Motivation
To keep consistency, it makes sense for Funding Sensitivity to be a WAD value. This way we can have very low funding sensitivity if needed and it allows us to keep precision in calcs.

# Changes
- make assumption funding sensitivity is wad and update calcs in pricing
- update comments
- remove PRB unsigned math from Pricing as its not used
- get rid of redundant function argument in pricing


sporejack (approved):
LGTM!

kumar-ish (approved):

PR #113, opened 2021-05-24
fix calculateTWAP and patch tcr-07

raymogg:
# Motivation
The current approach for calculating TWAPs was incorrectly implemented as it wasn't factoring in the `hour` parameter. Also took the opportunity to patch the bug with calculating the TWAP with only a single entry when updating internal parameters in the pricing contract.

# Changes
- twap accounts for `hour` param
- compute twap before incrementing hour in `pricing.sol`

### DIFF
--- a/contracts/lib/LibPrices.sol
+++ b/contracts/lib/LibPrices.sol
@@ -101,7 +101,9 @@ library Prices {
 
         for (uint256 i = 0; i < 8; i++) {
             uint256 currTimeWeight = 8 - i;
-            uint256 j = hour < i ? 23 : hour - i;
+            // if hour < i loop back towards 0 from 23.
+            // otherwise move from hour towards 0
+            uint256 j = hour < i ? 23 - i + hour : hour - i;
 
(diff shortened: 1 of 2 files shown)


contracts/lib/LibPrices.sol
```diff
@@ -92,7 +101,7 @@ library Prices {
 
         for (uint256 i = 0; i < 8; i++) {
             uint256 currTimeWeight = 8 - i;
-            uint256 j = 8 - i;
+            uint256 j = hour < i ? 23 : hour - i;
```
CalabashSquash: If, for example, `hour == 2`, then when i is from 3 -> 7, `j` is going to be 23 for all of those iterations.

Is this OK?

raymogg: @BenjaminPatch good catch. This is actually a bug in the old implementation that nobody caught 👀 

It should be continue looping towards 0 rather than repeating 23. Will fix

CalabashSquash (approved):
:st_kitts_nevis:

sporejack (approved):
LGTM!

PR #114, opened 2021-05-24
Unit test TracerPerpetualSwaps

raymogg:
# Motivation
The perp is the core of all our contracts and needs adequate unit testing (outside of our library unit testing)


test/unit/TracerPerpetualSwaps.js
```diff
@@ -0,0 +1,267 @@
...
+const setup = deployments.createFixture(async () => {
+    // todo setup and mock appropriate functions
+})
+
+describe("Unit tests: Insurance.sol", function () {
+    beforeEach(async function () {
+        // todo call setup
+    })
+
+    describe("deposit", async () => {
```
CalabashSquash, line 81: What about testing `deposit` when the amount is too high (i.e. `amount` WADified goes over `MAX_INT256`)?

raymogg: @BenjaminPatch I'm thinking this might be part of unit tests for LibBalances? Since it should test that those functions revert on overflow

contracts/TracerPerpetualSwaps.sol
```diff
@@ -696,7 +688,7 @@ contract TracerPerpetualSwaps is
     modifier onlyLiquidation() {
         require(
             msg.sender == liquidationContract,
-            "TCR: Sender not liquidation contract "
+            "TCR: Sender not liquidation"
         );
         _;
```
CalabashSquash, line 693: I feel like this makes it less clear but not too much of an issue

contracts/lib/LibBalances.sol
```diff
@@ -15,24 +15,33 @@ library Balances {
...
     }
 
+    /**
+     * @notice Calculates the net value of a position as base * price
+     * @param position the position the account is currently in
+     * @param price The (fair) price of the base asset
+     * @return Net value of a position given the price
+     */
     function netValue(Position memory position, uint256 price)
         internal
```
CalabashSquash: I reckon we just change this back to notionalValue now to be fair

contracts/lib/LibBalances.sol
```diff
@@ -94,10 +104,22 @@ library Balances {
         }
     }
 
+    /**
+     * @notice Calculates the minimum margin needed for an account.
+     * Calculated as minMargin = netValue / maxLev + liquidationGasCost
+     *                         = (base * price) / maxLev + liquidationGasCost
```
CalabashSquash:
```suggestion
     * Calculated as minMargin = notionalValue / maxLev + 6 * liquidationGasCost
     *                         = (base * price) / maxLev + 6 * liquidationGasCost
```

contracts/lib/LibBalances.sol
```diff
@@ -94,10 +104,22 @@ library Balances {
...
+     * @param position Position to calculate the minimum margin for
+     * @param price Price by which to evaluate the minimum margin
+     * @param liquidationGasCost Cost for liquidation
```
CalabashSquash:
```suggestion
     * @param liquidationGasCost Cost for liquidation denominated in quote tokens
```

CalabashSquash (changes requested):

contracts/TracerPerpetualSwaps.sol
raymogg: Yeah valid @BenjaminPatch , was just doing this for gas optimisation. "Require strings
If we are adding message strings to require statements, we can make them cheaper by limiting the string length to 32 bytes." https://medium.com/coinmonks/gas-optimization-in-solidity-part-i-variables-9d5775e43dde

contracts/lib/LibBalances.sol
raymogg: @BenjaminPatch agreed, we refer to it as notional value everywhere else so may as well

CalabashSquash (approved):
:malawi:

PR #115, opened 2021-05-25
Use `LibBalances` throughout contracts

sporejack:
# Motivation
The recent refactor focused primarily on extracting logic into separate libraries. While this has progressed well, some of the actual contracts themselves likely contain artifacts of the previous, non-library implementation. These need to be changed.

# Changes
 - Refactor `executeTrade` to use the corresponding `LibBalances.applyTrade` (which was written for this exact purpose)
 - Remove the (largely redundant) `fillAmount` parameter from the external interface
 - Extract margin validation into `LibBalances`
 - Extract funding rate calculations into `LibPrices`
 - Extract insurance calculations into `LibPrices`

### DIFF
--- a/contracts/TracerPerpetualSwaps.sol
+++ b/contracts/TracerPerpetualSwaps.sol
@@ -213,27 +213,27 @@ contract TracerPerpetualSwaps is
         bytes32 orderId1 = Perpetuals.orderId(order1);
         bytes32 orderId2 = Perpetuals.orderId(order2);
 
+        uint256 fillAmount =
+            Balances.fillAmount(
+                trade1,
+                filled[orderId1],
+                trade2,
+                filled[orderId2]
+            );
+
         // Calculate new account state
         (Balances.Position memory newPos1, Balances.Position memory newPos2) =
             (
                 Balances.applyTrade(
                     account1.position,
                     trade1,
-                    Balances.fillAmount(
-                        trade1,
-                        filled[orderId1],
-                        filled[orderId2]
-                    ),
+                    fillAmount,
                     feeRate
                 ),
                 Balances.applyTrade(
                     account2.position,
                     trade2,
-                    Balances.fillAmount(
-                        trade2,
-                        filled[orderId1],
-                        filled[orderId2]
-                    ),
+                    fillAmount,
                     feeRate
                 )
             );
@@ -243,7 +243,6 @@ contract TracerPerpetualSwaps is
         account2.position = newPos2;
 
         // Add fee into cumulative fees
-        uint256 fillAmount = LibMath.min(order1.amount, order2.amount);
         int256 quoteChange =
             PRBMathUD60x18.mul(fillAmount, order1.price).toInt256();
         int256 fee =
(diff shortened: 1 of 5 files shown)


contracts/TracerPerpetualSwaps.sol
```diff
@@ -198,55 +197,38 @@ contract TracerPerpetualSwaps is
...
+        (Balances.Trade memory trade1, Balances.Trade memory trade2) =
+            (
+                Balances.Trade(order1.price, order1.amount, order1.side),
+                Balances.Trade(order2.price, order2.amount, order2.side)
+            );
+
+        // Calculate new account state
+        (Balances.Position memory newPos1, Balances.Position memory newPos2) =
+            (
+                Balances.applyTrade(account1.position, trade1, feeRate),
```
raymogg: `applyTrade` applies the entire trade amount to the position, not the match amount. Some how we need to either factor `fillAmount` into this or alter the actual trade object to have the correct amount

contracts/lib/LibBalances.sol
```diff
@@ -33,7 +33,7 @@ library Balances {
         uint256 lastUpdatedGasPrice;
     }
 
-    function netValue(Position calldata position, uint256 price)
+    function netValue(Position memory position, uint256 price)
         public
```
CalabashSquash, line 37: not related to this PR, but we should make these functions `internal`

raymogg (changes requested):

contracts/lib/LibBalances.sol
jmcph4: Yeah there's a ticket for it - was going to be my next one!

sporejack: There's a ticket for this, I'll make it my next PR.

contracts/TracerPerpetualSwaps.sol
sporejack: Consult c1c2220

contracts/TracerPerpetualSwaps.sol
```diff
@@ -198,55 +197,59 @@ contract TracerPerpetualSwaps is
...
+                    trade2,
+                    Balances.fillAmount(
+                        trade2,
+                        filled[orderId1],
+                        filled[orderId2]
+                    ),
+                    feeRate
+                )
+            );
+
```
CalabashSquash: Fill amount should be the `min(order1.amount - filled[orderId1], order2.amount - filled[order2])`. As in, if order1 has 10 units left over, but order2 has 5 units left over, the fillAmount for both sides of the order should be 5 units.

At the moment, it makes the `fillAmount` of one side `min(order1.amount - filled[orderId1], order1.amount - filled[order2])`, and the `fillAmount` of the other side `min(order2.amount - filled[orderId1], order2.amount - filled[order2])`

CalabashSquash (changes requested):

contracts/lib/LibBalances.sol
```diff
@@ -110,11 +110,22 @@ library Balances {
         return liquidationGasCost + minimumMarginWithoutGasCost;
     }
 
+    function fillAmount(
```
raymogg, line 113: I reckon fillAmount should just take in both orders and both an aFilled and bFilled and return. That way we can call it once from the perp

sporejack: Consult d0486d1

raymogg (approved):

PR #116, opened 2021-05-25
Switch liquidation receipts to use enum Side

CalabashSquash:
### Motivation
Now that we are using an `Enum` to tell which side an order is (long vs short), we should use this in `LibLiquidation.LiquidationReceipt`

### Changes
- Switched all references to `liquidationSide` to use the `Enum Side` in `LibPerpetuals.sol`

### DIFF
--- a/contracts/test/LibLiquidationMock.sol
+++ b/contracts/test/LibLiquidationMock.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.0;
 
 import "../lib/LibLiquidation.sol";
+import "../lib/LibPerpetuals.sol";
 
 library LibLiquidationMock {
     function calcEscrowLiquidationAmount(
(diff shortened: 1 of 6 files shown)


contracts/test/LibLiquidationMock.sol
```diff
@@ -45,7 +46,7 @@ library LibLiquidationMock {
         uint256 maxSlippage,
         uint256 avgPrice,
         uint256 receiptPrice,
-        bool receiptSide
+        uint256 receiptSide
```
raymogg, line 49: any specific reason for doing this over just passing in a `Perpetuals.Side` enum?

raymogg (approved):

contracts/test/LibLiquidationMock.sol
CalabashSquash: Yeah there was a type mismatch if I tried to make it `Perpetuals.Side`, but it worked if I casted inside the contracts

PR #117, opened 2021-05-25
Consistent withdraw and deposit

raymogg:
# Motivation
Previously deposit took in `amount` as a raw token amount, and withdraw as a WAD amount. To keep this consistent, these have all been changed to WAD amounts.

# Changes
- change `amount` on deposit to be a WAD
- rename `stake` to `deposit` on `insurance.sol`

### DIFF
--- a/contracts/Insurance.sol
+++ b/contracts/Insurance.sol
@@ -49,11 +49,11 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
     }
 
     /**
-     * @notice Allows a user to stake to a given tracer market insurance pool
+     * @notice Allows a user to deposit to a given tracer market insurance pool
      * @dev Mints amount of the pool token to the user
      * @param amount the amount of tokens to stake. Provided in WAD format
      */
-    function stake(uint256 amount) external override {
+    function deposit(uint256 amount) external override {
         IERC20 collateralToken = IERC20(collateralAsset);
         collateralToken.transferFrom(msg.sender, address(this), amount);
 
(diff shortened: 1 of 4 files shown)


CalabashSquash (approved):

contracts/Insurance.sol
```diff
@@ -49,17 +49,17 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
...
-     * @notice Allows a user to stake to a given tracer market insurance pool
+     * @notice Allows a user to deposit to a given tracer market insurance pool
      * @dev Mints amount of the pool token to the user
-     * @param amount the amount of tokens to stake. Provided in token native units
+     * @param amount the amount of tokens to stake. Provided in WAD format
      */
-    function stake(uint256 amount) external override {
+    function deposit(uint256 amount) external override {
         IERC20 collateralToken = IERC20(collateralAsset);
         collateralToken.transferFrom(msg.sender, address(this), amount);
```
CalabashSquash: This should be un-wadified, right?

CalabashSquash (changes requested):

contracts/Insurance.sol
raymogg: ah yeah I need to apply the same logic here as I did in the perp itself.

contracts/TracerPerpetualSwaps.sol
```diff
@@ -95,27 +95,35 @@ contract TracerPerpetualSwaps is
...
             address(this),
-            amount
+            amountToTransfer
         );
 
+        // this prevents dust from being added to the user account
+        // eg 10^18 -> 10^8 -> 10^18 will remove lower order bits
+        int256 amountFormatted =
+            Balances.tokenToWad(quoteTokenDecimals, amountToTransfer);
+
```
CalabashSquash: This is smart :100: :kiribati:

CalabashSquash (approved):
pending lint

PR #118, opened 2021-05-25
Margin is valid with no base

raymogg:
# Motivation
There was a bug with how withdraw was computing `marginIsValid`. This allowed a user to withdraw more than deposited when they had no base (positions) outstanding.

# Changes
- ensure quote > 0 if base = 0 in `marginIsValid`
- removes redundant `margin < 0` check

### DIFF
--- a/contracts/TracerPerpetualSwaps.sol
+++ b/contracts/TracerPerpetualSwaps.sol
@@ -453,13 +453,10 @@ contract TracerPerpetualSwaps is
             Balances.minimumMargin(pos, price, gasCost, maxLeverage);
         int256 margin = Balances.margin(pos, price);
 
-        if (margin < 0) {
-            /* Margin being less than 0 is always invalid, even if position is 0.
-               This could happen if user attempts to over-withdraw */
-            return false;
-        }
         if (minMargin == 0) {
-            return true;
+            // minMargin = 0 only occurs when user has no base (positions)
+            // if they have no base, their quote must be > 0.
+            return quote >= 0;
         }
 
         // todo CASTING CHECK
--- a/contracts/TracerPerpetualSwaps.sol
+++ b/contracts/TracerPerpetualSwaps.sol
@@ -460,6 +460,12 @@ contract TracerPerpetualSwaps is
         uint256 minMargin =
             Balances.minimumMargin(pos, price, gasCost, maxLeverage);
         int256 margin = Balances.margin(pos, price);
+        
+        if (margin < 0) {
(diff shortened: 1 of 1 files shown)


contracts/TracerPerpetualSwaps.sol
```diff
@@ -453,13 +453,10 @@ contract TracerPerpetualSwaps is
             Balances.minimumMargin(pos, price, gasCost, maxLeverage);
         int256 margin = Balances.margin(pos, price);
 
-        if (margin < 0) {
-            /* Margin being less than 0 is always invalid, even if position is 0.
-               This could happen if user attempts to over-withdraw */
-            return false;
-        }
```
CalabashSquash: Shouldn't this still be in anyway?

contracts/TracerPerpetualSwaps.sol
```diff
@@ -453,13 +453,10 @@ contract TracerPerpetualSwaps is
...
         if (minMargin == 0) {
-            return true;
+            // minMargin = 0 only occurs when user has no base (positions)
+            // if they have no base, their quote must be > 0.
```
sporejack, line 494: Mismatch between comment and actual statement below. I'm assuming the comment is wrong and should be inclusive?

sporejack: I.e.,

```suggestion
            // if they have no base, their quote must be >= 0.
```

contracts/TracerPerpetualSwaps.sol
raymogg: @BenjaminPatch `Balances.minimumMargin` now returns a uint so it cant be less than 0. I think this existed as previously we returned a int and then casted so we needed to make this safety check.

CalabashSquash: This is margin though, not minimum margin

sporejack (changes requested):
Might be a controversial opinion, but should we just convert the body of `marginIsValid` into a ternary?

contracts/TracerPerpetualSwaps.sol
raymogg: @sporejack are you giving me permission to make a triple chained ternary? 👀

CalabashSquash (approved):
:sri_lanka:

contracts/TracerPerpetualSwaps.sol
sporejack: @raymogg Not anymore!

PR #119, opened 2021-05-26
Simulate trades

dospore:
### Motivation
- need to test recording users trades

### Changes
- Add another script which runs a bunch of trades between acc1 and acc2 
- Change fullDeploy so it deploys through the factory and not through hardhat

### DIFF
--- a/deploy/FullDeploy.js
+++ b/deploy/FullDeploy.js
@@ -45,19 +45,6 @@ module.exports = async function (hre) {
     })
 
     // deploy oracles
-    const priceOracle = await deploy("PriceOracle", {
-        from: deployer,
-        log: true,
-        contract: "Oracle",
-    })
-
-    const oracleAdapter = await deploy("PriceOracleAdapter", {
-        from: deployer,
-        log: true,
-        args: [priceOracle.address],
-        contract: "OracleAdapter",
-    })
-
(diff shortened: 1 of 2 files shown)


CalabashSquash (approved):

deploy/FullDeploy.js
```diff
@@ -45,29 +45,23 @@ module.exports = async function (hre) {
...
     })
 
-    const ethOracle = await deploy("EthOracle", {
+    const oracleAdapter = await deploy("PriceOracleAdapter", {
         from: deployer,
         log: true,
-        contract: "Oracle",
+        args: [ethOracle.address],
+        contract: "OracleAdapter",
     })
```
CalabashSquash: This is using ethOracle, but it should be using priceOracle, which you've deleted.

price Oracle is for the price of the base asset, eth oracle is for ETH/USD (used for calculating liquidation gas cost)

CalabashSquash (changes requested):

deploy/FullDeploy.js
dospore: is see, but this is working now. Should the priceOracle be passed into the Tracer contract??

CalabashSquash (approved):
:fiji:

PR #121, opened 2021-05-26
Pricing Unit Tests

raymogg:
# Motivation
Adds unit tests for the Pricing library

# Changes
- unit tests for pricing
- alters the `globalLeverage` function implementation
- alters `twap` implementation
-


contracts/lib/LibPrices.sol
```diff
@@ -36,14 +36,19 @@ library Prices {
...
         public
         pure
         returns (uint256)
     {
+        // todo double check safety of this.
+        // average price == 0 is not neccesarily the
+        // same as no trades in average
         if (price.trades == 0) {
             return 0;
         }
```
CalabashSquash, line 61: Yeah, we need to make sure that having the contract consider averagePrice to be 0 on no trades is safe. Could this mess up the funding rate towards the start of a market's deployment?

raymogg: @BenjaminPatch I'll have a look into this now. Are you happy for me to do this as part of this PR if there are changes to be made?

CalabashSquash: Yeah I am

contracts/lib/LibPrices.sol
```diff
@@ -94,45 +104,28 @@ library Prices {
...
+            totalTimeWeight += currTimeWeight;
+            cumulativeDerivative += currTimeWeight * currDerivativePrice;
+            cumulativeUnderlying += currTimeWeight * currUnderlyingPrice;
         }
+
+        return
+            TWAP(
+                cumulativeUnderlying / totalTimeWeight,
+                cumulativeDerivative / totalTimeWeight
+            );
```
CalabashSquash: Should this use `PRBMath.div`?

raymogg: Nah I changed this since it actually shouldn't have been. Using WAD maths would mean that it interprets totalTimeWeight as a WAD value and would ensure the result is a WAD, however we want it to actually decrease decimal places if it needs to (which is the opposite of WAD).

CalabashSquash (approved):

test/unit/LibPrices.js
```diff
@@ -0,0 +1,393 @@
...
+                    ])
+                    tracerPrices.push([
+                        ethers.utils.parseUnits((1 + 0.5 * i).toString(), 18),
+                        ethers.BigNumber.from("1"),
+                    ])
+                }
+
+                // generate expected results
+                expectedTwaps = new Array()
+                for (var hour = 0; hour < 23; hour++) {
```
kumar-ish:
```suggestion
                for (var hour = 0; hour < 24; hour++) {
```
Should be `24`

test/unit/LibPrices.js
```diff
@@ -0,0 +1,393 @@
...
+                    ])
+                    tracerPrices.push([
+                        ethers.utils.parseEther("0"),
+                        ethers.BigNumber.from("1"),
+                    ])
+                }
+
+                // generate expected results
+                expectedTwaps = new Array()
+                for (var hour = 0; hour < 23; hour++) {
```
kumar-ish:
```suggestion
                for (var hour = 0; hour < 24; hour++) {
```
Should be `24`

contracts/lib/LibPrices.sol
```diff
@@ -55,12 +60,19 @@ library Prices {
...
-        uint256 n = prices.length <= 24 ? prices.length : 24;
+        uint256 n = (prices.length <= 24) ? prices.length : 24;
         uint256[] memory averagePrices = new uint256[](24);
 
         for (uint256 i = 0; i < n; i++) {
             PriceInstant memory currPrice = prices[i];
-            averagePrices[i] = averagePrice(currPrice);
+            uint256 _averagePrice = averagePrice(currPrice);
+
+            // dont inclue periods that have no trades
```
kumar-ish:
```suggestion
            // dont include periods that have no trades
```

contracts/lib/LibPrices.sol
```diff
@@ -94,45 +111,41 @@ library Prices {
...
             // otherwise move from hour towards 0
-            uint256 j = hour < i ? 23 - i + hour : hour - i;
+            uint256 j = hour < i ? 24 - i + hour : hour - i;
 
             uint256 currDerivativePrice = averagePrice(tracerPrices[j]);
             uint256 currUnderlyingPrice = averagePrice(oraclePrices[j]);
 
-            if (currDerivativePrice > 0) {
-                instantDerivative += currTimeWeight;
+            // dont inclue periods that have no trades
```
kumar-ish:
```suggestion
            // dont include periods that have no trades
```

contracts/lib/LibPrices.sol
```diff
@@ -94,45 +111,41 @@ library Prices {
...
 
-            if (currUnderlyingPrice > 0) {
-                instantUnderlying += currTimeWeight;
-                cumulativeUnderlying += currTimeWeight * currUnderlyingPrice;
-            }
-
-            if (instantDerivative == 0) {
-                return TWAP(0, 0);
+            // dont inclue periods that have no trades
+            if (currUnderlyingPrice == 0 && oraclePrices[j].trades == 0) {
```
kumar-ish:
```suggestion
            if (currUnderlyingPrice == 0) {
```

`oraclePrices[j].trades` is always going to be the same as `tracerPrices[j].trades` so the latter check is redundant

contracts/lib/LibPrices.sol
```diff
@@ -94,45 +111,41 @@ library Prices {
...
             }
 
-            if (currUnderlyingPrice > 0) {
-                instantUnderlying += currTimeWeight;
-                cumulativeUnderlying += currTimeWeight * currUnderlyingPrice;
-            }
-
-            if (instantDerivative == 0) {
-                return TWAP(0, 0);
+            // dont inclue periods that have no trades
```
kumar-ish:
```suggestion
            // dont include periods that have no trades
```

kumar-ish (changes requested):
If you're in the mood for documentation, feel free to add it; otherwise I can later :surfer:

contracts/lib/LibPrices.sol
raymogg: If we consider each of these functions pure (eg can take any input) then this isn't strictly true. I'd rather be cautious than make use case specific assumptions.

contracts/lib/LibPrices.sol
```diff
@@ -96,46 +112,42 @@ library Prices {
...
             // otherwise move from hour towards 0
-            uint256 j = hour < i ? 23 - i + hour : hour - i;
+            uint256 j = hour < i ? 24 - i + hour : hour - i;
 
             uint256 currDerivativePrice = averagePrice(tracerPrices[j]);
             uint256 currUnderlyingPrice = averagePrice(oraclePrices[j]);
 
-            if (currDerivativePrice > 0) {
-                instantDerivative += currTimeWeight;
+            // dont include periods that have no trades
```
sporejack:
```suggestion
            // don't include periods that have no trades
```

contracts/lib/LibPrices.sol
```diff
@@ -96,46 +112,42 @@ library Prices {
...
             }
 
-            if (currUnderlyingPrice > 0) {
-                instantUnderlying += currTimeWeight;
-                cumulativeUnderlying += currTimeWeight * currUnderlyingPrice;
-            }
-
-            if (instantDerivative == 0) {
-                return TWAP(0, 0);
+            // dont include periods that have no trades
```
sporejack:
```suggestion
            // don't include periods that have no trades
```

sporejack (changes requested):

contracts/lib/LibPrices.sol
```diff
@@ -65,7 +65,14 @@ library Prices {
...
             PriceInstant memory currPrice = prices[i];
-            averagePrices[i] = averagePrice(currPrice);
+            uint256 _averagePrice = averagePrice(currPrice);
+
+            // dont inclue periods that have no trades
+            if (_averagePrice == 0 && currPrice.trades == 0) {
+                continue;
+            } else {
+                averagePrices[i] = averagePrice(currPrice);
+            }
```
BenjaminPatch: I think the `if... else...` is redundant here because in both the cases `_averagePrice == 0 && currPrice.trades == 0`, `averagePrice(...)` would have returned 0:
- https://github.com/tracer-protocol/tracer-protocol/blob/contracts-v2/contracts/lib/LibPrices.sol#L49

CalabashSquash: I think the if... else... is redundant here because in both the cases _averagePrice == 0 && currPrice.trades == 0, averagePrice(...) would have returned 0:

- https://github.com/tracer-protocol/tracer-protocol/blob/contracts-v2/contracts/lib/LibPrices.sol#L49

CalabashSquash (changes requested):

contracts/lib/LibPrices.sol
kumar-ish: This was a bit broken before; fixed now so that it ignores periods where there's zero trades in the mean

kumar-ish (approved):

kumar-ish:
Still need to add unit tests for some of the remaining functions in the LibPricing contract, and do some more edge cases for some of some of the other ones, but otherwise I think we should merge this sooner rather than later. @BenjaminPatch Review again?

CalabashSquash (approved):

contracts/lib/LibPrices.sol
```diff
@@ -38,53 +38,91 @@ library Prices {
...
     function averagePriceForPeriod(PriceInstant[24] memory prices)
         public
         pure
         returns (uint256)
     {
-        uint256 n = prices.length <= 24 ? prices.length : 24;
         uint256[] memory averagePrices = new uint256[](24);
 
-        for (uint256 i = 0; i < n; i++) {
+        // TODO: make sure this procedure is gas-optimised
```
sporejack, line 78: Is this something we still need to address?

contracts/lib/LibPrices.sol
```diff
@@ -96,48 +134,45 @@ library Prices {
...
         uint256 cumulativeUnderlying = 0;
 
         for (uint256 i = 0; i < 8; i++) {
             uint256 currTimeWeight = 8 - i;
             // if hour < i loop back towards 0 from 23.
             // otherwise move from hour towards 0
-            uint256 j = hour < i ? 23 - i + hour : hour - i;
+            uint256 j = hour < i ? 24 - i + hour : hour - i;
 
             uint256 currDerivativePrice = averagePrice(tracerPrices[j]);
```
sporejack, line 150: If `hour` is greater than or equal to 24, these two array accesses will be out-of-bounds, correct?

contracts/lib/LibPrices.sol
```diff
@@ -96,48 +134,45 @@ library Prices {
...
-                        )
-                    );
+                totalUnderlyingTimeWeight += currTimeWeight;
+                cumulativeUnderlying += currTimeWeight * currUnderlyingPrice;
             }
         }
+
+        return
+            TWAP(
+                cumulativeUnderlying / totalUnderlyingTimeWeight,
```
sporejack, line 172: I think this has come up before, but is there no need for WAD divisions here?

contracts/lib/LibPrices.sol
```diff
@@ -38,53 +38,91 @@ library Prices {
...
-
-        if (leverageHasIncreased) {
-            return _globalLeverage + (newLeverage - oldLeverage);
-        } else {
-            return _globalLeverage - (newLeverage - oldLeverage);
+        int256 newGlobalLeverage =
+            int256(_globalLeverage) +
+                (int256(newLeverage) - int256(oldLeverage));
+
+        // note: this would require a bug in how account leverage was recorded
```
sporejack, line 112: If global leverage is a stored contract variable, then this may be eligible for an Echidna predicate.

sporejack (changes requested):
Likely need an explicit bounds check on the `hour` parameter in `globalLeverage`, but of course, the comments are open for extended discussion here.

contracts/lib/LibPrices.sol
CalabashSquash: I don't think so. We could test if memory or storage for currPrice is cheaper but I think it's fine.

contracts/lib/LibPrices.sol
CalabashSquash: Yeah, how should this behave if `hour >= 24`? 
@kumar-ish

contracts/lib/LibPrices.sol
CalabashSquash: I would say it should be WAD

contracts/lib/LibPrices.sol
kumar-ish: It shouldn't be valid; I guess we can do an explicit check/require for it, or otherwise it'll just revert

contracts/lib/LibPrices.sol
kumar-ish: I just wrote it since the `continue`/second-increment variable was the first thing that came to mind, and probably could be more effecient somehow

contracts/lib/LibPrices.sol
CalabashSquash: Don't believe we can do a require because library functions should be pure but yep agreed

contracts/lib/LibPrices.sol
kumar-ish: Discussed in person; doesn't need to be WAD

contracts/lib/LibPrices.sol
kumar-ish: Added `require` for hour check

sporejack (approved):
LGTM!

PR #122, opened 2021-05-26
Make grabOrder return memory & rearrange LibPerpetuals.orderId signature

CalabashSquash:
### Motivation
grabOrder returns a storage struct. But given that it's not actually modified (Order types  are now immutable), it should copy it straight to memory for gas savings.

Also, while writing tests on a separate branch, I had some issues with getting getOrder to work and found that with some rearranging it was able to be fixed.

### Changes
- Make `Trader.grabOrder(...)` return memory struct, not storage
- Make `Perpetuals.orderId(...)` take memory struct, and make it internal
- Add orderId to LibPerpetualsMock
- Change contract name `libPerpetualsMock -> PerpetualsMock` to be consistent with `Perpetuals`.

### DIFF
--- a/contracts/lib/LibPerpetuals.sol
+++ b/contracts/lib/LibPerpetuals.sol
@@ -14,7 +14,7 @@ library Perpetuals {
         uint256 created;
     }
 
-    function orderId(Order calldata order) public pure returns (bytes32) {
+    function orderId(Order memory order) internal pure returns (bytes32) {
         return keccak256(abi.encode(order));
     }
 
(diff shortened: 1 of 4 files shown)


raymogg (approved):
🏄‍♂️

contracts/lib/LibPerpetuals.sol
```diff
@@ -14,7 +14,7 @@ library Perpetuals {
         uint256 created;
     }
 
-    function orderId(Order calldata order) public pure returns (bytes32) {
+    function orderId(Order memory order) internal pure returns (bytes32) {
```
sporejack, line 17: Will we ever want external callers to be able to calculate unique identifiers for orders?

sporejack (approved):

PR #123, opened 2021-05-27
Added check for order price in calcUnitsSold

CalabashSquash:
### Motivation
If an order is given during slippage submit, and the order has negative slippage (price up when liquidation was long, vise-versa), then it should be ignored

### Changes
- Ignore orders that had negative slippage on calcUnitsSold


sporejack (approved):
:car:

PR #124, opened 2021-05-27
Deployments

raymogg:
# Motivation
In order to start deploying to test networks, our deployment script needed to log more information. I also removed some redundant code in `DeployAndFund` as `FullDeploy` already did this.

# Changes
- logs
- don't re deploy insurance, pricing and liquidation
- remove code from `DeployAndFund`

### DIFF
--- a/deploy/FullDeploy.js
+++ b/deploy/FullDeploy.js
@@ -96,7 +96,9 @@ module.exports = async function (hre) {
         contract: "OracleAdapter",
     })
 
-    console.log(`Gas Price Oracle Adapter Deployed ${gasPriceOracleAdapter.address}`)
+    console.log(
+        `Gas Price Oracle Adapter Deployed ${gasPriceOracleAdapter.address}`
+    )
 
     // deploy token with an initial supply of 100000
     const token = await deploy("QuoteToken", {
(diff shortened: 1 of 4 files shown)


deploy/FullDeploy.js
```diff
@@ -165,8 +193,10 @@ module.exports = async function (hre) {
         log: true,
     })
 
+    console.log(`Factory Deployed: ${factory.address}`)
+
     let maxLeverage = ethers.utils.parseEther("12.5")
-    let tokenDecimals = new ethers.BigNumber.from("8").toString()
+    let tokenDecimals = new ethers.BigNumber.from("18").toString()
```
CalabashSquash, line 199: Is this correct? I think the quote token hasn't been modified from its default 8 decimals.

CalabashSquash (changes requested):

deploy/FullDeploy.js
raymogg: @BenjaminPatch the quote token (`TestToken.sol`) doesn't take in a decimals option so it defaults to 18 I am pretty sure. We also pass in the totalSupply as a 18 decimal number

dospore (approved):

deploy/FullDeploy.js
CalabashSquash: Yep that makes sense. I was getting this mixed up with the price oracle.

CalabashSquash (approved):

PR #125, opened 2021-05-27
Mark `LibBalances` functions as internal

sporejack:
# Motivation
Library functions should be `internal`.

# Changes
 - Change the visibility of all functions in the `LibBalances` library from `public` to `internal`
 - Remove explicit linking from `Insurance.sol` tests

### DIFF
--- a/test/unit/Insurance.js
+++ b/test/unit/Insurance.js
@@ -35,7 +35,6 @@ const setup = deployments.createFixture(async () => {
         "TracerPerpetualSwaps",
         {
             libraries: {
-                Balances: libBalances.address,
                 Perpetuals: libPerpetuals.address,
                 Prices: libPrices.address,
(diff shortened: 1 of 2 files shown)


CalabashSquash (approved):

test/unit/Insurance.js
```diff
@@ -35,7 +35,6 @@ const setup = deployments.createFixture(async () => {
         "TracerPerpetualSwaps",
         {
             libraries: {
-                Balances: libBalances.address,
```
raymogg, line 38: Doesn't the perp need LibBalances?

sporejack: It does, but @BenjaminPatch and I came to the (verbal) conclusion that you mustn't need to explicitly link `internal` libraries.

PR #126, opened 2021-05-27
Remove instances of boolean side representation

sporejack:
# Motivation
Market sides have a dedicated enumerated type now -- we should use them.

# Changes
 - Change liquidation events to use the `Perpetuals.Side` type


raymogg (approved):

CalabashSquash (approved):
:myanmar:

PR #130, opened 2021-06-01
Add insurance buffer

kumar-ish:
## Motivation

As per the Tracer Perpetual Swaps whitepaper, the insurance pool for a market is composed of two accounts: the system-owned buffer insurance account, and the public insurance account. This pull request implements the system-owned buffer insurance account.

## Changes
* Add insurance buffer account

* Change respective functions to account for the insurance buffer

Note: Scope of PR changed so as to not include the insurance withdrawal fee, as aspects of it are still under research and development


contracts/Insurance.sol
```diff
@@ -68,11 +69,11 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
         updatePoolAmount();
         InsurancePoolToken poolToken = InsurancePoolToken(token);
 
-        // tokens to mint = (pool token supply / collateral holdings) * collaterael amount to stake
+        // tokens to mint = (pool token supply / collateral holdings) * collateral amount to stake
```
greenbookwebb, line 62: How are we defining collateral holdings? Is collateral holdings just collateral amount (public), or is it collateral amount (public) + buffer? 

It should be the second one

contracts/Insurance.sol
```diff
@@ -99,21 +100,26 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
         uint256 wadTokensToSend =
             LibInsurance.calcWithdrawAmount(
```
greenbookwebb: Just confiming that this function only allows you to withdraw from publc (collateralAmount), and not from bufferAmount.

contracts/Insurance.sol
```diff
@@ -124,7 +130,12 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
         int256 quote = (tracer.getBalance(address(this))).position.quote;
         if (quote > 0) {
             tracer.withdraw(uint256(quote));
-            collateralAmount = collateralAmount + uint256(quote);
+            if (collateralAmount > 0) {
+                collateralAmount = collateralAmount + uint256(quote);
+            } else {
+                // Pay to buffer if nothing in public insurance
+                bufferAmount = bufferAmount + uint256(quote);
+            }
```
greenbookwebb: Some of this is true, some is incorrect. 

Insurance funding rate payments are paid proportionally to the two pools. e.g. If insurance funding payment = 25 Quote, Insurance buffer account has 10 units of quote already in the pool (10% of total), public account has 90 units of quote (90% of total), Insurance buffer account increases by 25*0.1, and public account increases by 25*0.9. 

BUT, if neither pool has any quote collateral in their account (like maybe at genesis), it will go 100% to insurance buffer account.

contracts/Insurance.sol
```diff
@@ -199,7 +214,7 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
 
         uint256 ratio =
             PRBMathUD60x18.div(
-                getPoolTarget() - collateralAmount,
+                getPoolTarget() - collateralAmount - bufferAmount,
```
greenbookwebb: unsure what this does.

greenbookwebb (changes requested):

contracts/Insurance.sol
CalabashSquash: I think, at the moment, collateralAmount is just the public pool. However, it could probably be renamed to publicCollateralAmount.

contracts/Insurance.sol
CalabashSquash: Yeah, I think `calcWithdrawAmount` should only take public collat amount, as opposed to public + buffer, right?

contracts/Insurance.sol
kumar-ish: Yep, that makes sense; will fix this up

CalabashSquash: Why do you say 250.1 and 250.9? Where have these numbers come from? EDIT: this is a markdown formatting thing lol

kumar-ish: Yep, that makes sense; can fix

contracts/Insurance.sol
kumar-ish: This ratio is `(insurance_fund_target - insurance_fund_holdings)/(leveraged_notional_value)` in the formula

kumar-ish: `collateralAmount` is the amount of quote in the public insurance pool and `bufferAmount` is the amount of quote in the buffer insurance pool

contracts/Insurance.sol
CalabashSquash: According to @kumar-ish , the buffer account only gets paid when
a) The public account is empty
b) People pay withdrawal fees

But you're saying it's proportional to pool holdings.

If your statement is correct, should this split be done as soon as insurance funding rate payments are made, or is it OK to make this calculation when `updatePoolAmount(...)` is called?

contracts/Insurance.sol
CalabashSquash: This makes the calculation for funding rate equal to

`36523 * (10^11) * [(target - public collat amount - buffer collat amount) / leveraged notional value]`

contracts/Insurance.sol
kumar-ish: This should be fixed now; also, discussed in-person with Ben and it should be okay to make the calculation when `updatePoolAmount()` is called

contracts/Insurance.sol
kumar-ish: Ooh, you're definitely right; fixed now

contracts/Insurance.sol
CalabashSquash: Yep it is, because `updatePoolAmount()` is called on any change to insurance pool balance. I was worried the ratios could be "outdated" by the time it is called, but this is not the case.

contracts/Insurance.sol
```diff
@@ -142,24 +163,28 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
...
-        // What the balance will be after
-        uint256 difference = collateralAmount - amount;
-        if (difference < 10**18) {
+        if (bufferAmount + collateralAmount < 10**18) {
             // Once we go below one token, social loss is required
             // This calculation caps draining so pool always has at least one token
-            amount = collateralAmount - (10**18);
-            // Use new amount to compute difference again.
-            difference = collateralAmount - amount;
+            bufferAmount = 10**18;
```
CalabashSquash: I think this is wrong. You can't just arbitrarily set bufferAmount to 1. Where is this token coming from?

CalabashSquash (changes requested):

contracts/Insurance.sol
kumar-ish: This is making the making sure there is at least one token in the insurance pool, but it should probably be a part of the public pool. I'm gonna speak to Webb about this and see how it should be specc'd

greenbookwebb: yep, as discussed, get rid - no need for this. Only for public account

contracts/Insurance.sol
```diff
@@ -199,7 +224,7 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
 
         uint256 ratio =
             PRBMathUD60x18.div(
-                getPoolTarget() - collateralAmount,
+                getPoolTarget() - collateralAmount - bufferAmount,
                 levNotionalValue
```
greenbookwebb: If `levNotionalValue == 0`, this will throw. This can happen naturally, or happen when `maxlev == 1`

greenbookwebb: @BenjaminPatch @kumar-ish

kumar-ish: There's a check for `levNotionalValue` on line 224; following block of code:
```
    if (levNotionalValue <= 0) {
        return 0;
    }
```

contracts/Insurance.sol
greenbookwebb: What's this for though? Insurance Funding Rate?

greenbookwebb (approved):

contracts/Insurance.sol
kumar-ish: Yep, that's right

contracts/Insurance.sol
kumar-ish: Fixed (I think)

contracts/Insurance.sol
```diff
@@ -137,29 +161,55 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
...
+            } else {
+                //
+                amount = bufferCollateralAmount;
+            }
+
+            // Drain buffer
+            bufferCollateralAmount = 0;
+        } else if (amount > bufferCollateralAmount) {
+            uint256 leftAfterDrain =
+                amount - bufferCollateralAmount - publicCollateralAmount;
```
CalabashSquash: Consider the example where
`publicCollateralAmount = 10`
`bufferCollateralAmount = 20`
`amount = 25`

then `amount > bufferCollateralAmount == true`
so `leftAfterDrain = 25 - 20 - 10` so underflow would occur.

It should make `leftAfterDrain = amount - bufferCollateralAmount`, then check if `amount > public`, if so drain all but 1. Then `amount = amountTakenFromBuffer + amountTakenFromPublic`, and set bufferCollateralAmount to =, and amountTakenFromPublic to be whatever it should be (min 1 token)

CalabashSquash (changes requested):

contracts/Insurance.sol
kumar-ish: Yes, you're right -- I believe the logic should be all _patched_ up now. I did some [tests](https://gist.github.com/kumar-ish/23ff05e1cdf52be6826e0f7ee4f500c0) in Python too, but feel free to attack it with edge cases

contracts/Insurance.sol
```diff
@@ -68,17 +69,17 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
...
                 poolToken.totalSupply(),
-                collateralAmount,
+                publicCollateralAmount,
                 wadAmount
             );
 
         // mint pool tokens, hold collateral tokens
         poolToken.mint(msg.sender, tokensToMint);
-        collateralAmount = collateralAmount + wadAmount;
+        publicCollateralAmount = publicCollateralAmount + wadAmount;
```
greenbookwebb, line 67: What is wadAmount? Is it the collateral that is being deposited into the pool when a user mints pool tokens?

kumar-ish: Yes, that's right

contracts/Insurance.sol
```diff
@@ -121,10 +123,32 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
...
+            publicCollateralAmount = publicCollateralAmount + payToCollateral;
+
+            // Amount to pay to buffer is the remainder
+            bufferCollateralAmount =
+                bufferCollateralAmount +
+                quote -
+                payToCollateral;
+        } else {
+            // Pay to buffer if nothing in public insurance
+            bufferCollateralAmount = bufferCollateralAmount + quote;
```
greenbookwebb: This doesn't look right. At the time a user withdraws from the public fund, the withdrawal fee goes entirely to the insurance buffer account.

contracts/Insurance.sol
```diff
@@ -137,29 +161,46 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
...
+                // from the public insurance such that 1 token is left in the public buffer
+                amount =
+                    bufferCollateralAmount +
+                    publicCollateralAmount - 
+                    10**18;
+                publicCollateralAmount = 10**18;
+            } else {
+                // Take out what you need from the public pool; there's enough for there to be >= 1 token left
+                publicCollateralAmount = poolHoldings - amount;
+            }
```
greenbookwebb: this looks correct, but there's a lot going on here so I won't confirm - @BenjaminPatch to confirm

greenbookwebb (changes requested):
Most of this looks good. Just got to fix up that bit with the withdrawal fee going entirely to the insurance buffer account.

contracts/Insurance.sol
```diff
@@ -121,10 +123,32 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
...
-        int256 quote = (tracer.getBalance(address(this))).position.quote;
-        if (quote > 0) {
-            tracer.withdraw(uint256(quote));
-            collateralAmount = collateralAmount + uint256(quote);
+        uint256 quote =
+            uint256((tracer.getBalance(address(this))).position.quote);
+
+        tracer.withdraw(quote);
+
+        if (publicCollateralAmount > 0) {
```
CalabashSquash, line 113: Just confirming this is correct, and shouldn't be `bufferCollateralAmount`.

kumar-ish: Yes; only when someone has deposited into the public insurance account do funds start accruing proportionally to the buffer and public pools; otherwise, they all go to the public pool

contracts/Insurance.sol
CalabashSquash: There is a lot going on but there are a decent amount of edge cases that need checking. I'm confident that this is good.

contracts/Insurance.sol
kumar-ish: This is for updating the amount of buffer and public collateral tokens in the insurance pool. In order to optimise for gas, all insurance funding payments are made in the `TracerPerpetualSwaps` contract; when you actually need to deposit or withdraw from the insurance pool, you need access to those collateral tokens, so `updatePoolAmount` gets them from the perp swaps contract (which is why `tracer.withdraw(quote);` is called at the beginning). 

```
            // Pay to buffer if nothing in public insurance
```
is saying that if no one has deposited collateral into the insurance account, then any amount that the Tracer has paid out to the insurance contract (from insurance funding rate payments) belongs to the buffer account. If someone has (i.e. the if-statement branch, where the collateral in the insurance contract is greater than 0), you pay proportionally to the buffer and public account.

The withdraw fee is still under discussion so I decided to leave it out of scope for this PR.

CalabashSquash (approved):

raymogg (approved):

PR #131, opened 2021-06-07
Error events for matching

raymogg:
# Motivation
In order to simplify frontend error handling, events are now emitted for failed orders. New matchOrders logic was added to do this, however none of the functionality should change

# Changes
- emit failed order event
- rearrange logic of matchOrders to return a bool and not revert.


contracts/lib/LibBalances.sol
```diff
@@ -155,6 +153,20 @@ library Balances {
         return newPosition;
     }
 
+    function getFeeRate(
+        uint256 amount,
+        uint256 executionPrice,
+        uint256 feeRate
+    ) internal pure returns (int256) {
```
BenjaminPatch, line 160: nitpick: I think this should be changed to `getFee` or `calculateFee`, since it uses the "rate" to calculate the absolute amount

CalabashSquash (approved):

contracts/lib/LibBalances.sol
CalabashSquash: I think this should be changed to getFee or calculateFee, since it uses the "rate" to calculate the absolute amount

CalabashSquash (changes requested):

sporejack (approved):

CalabashSquash (approved):
:timor_leste:

PR #132, opened 2021-06-08
Change execution price

raymogg:
# Motivation
Previously we were ensuring that the price of two orders was exactly the same for execution to happen. This however isn't the case. Orders should be executed if their prices cross. A long order price should be greater than or equal to the price of the short order.

# Changes
- adds a `getExecutionPrice` function
- use the execution price in perp
- alter the `canMatch` check to not enforce strict equals
- add new test


contracts/lib/LibPerpetuals.sol
```diff
@@ -113,4 +114,16 @@ library Perpetuals {
             notFilled &&
             createdBefore;
     }
+
+    function getExecutionPrice(
+        Order calldata a,
+        Order calldata b
+    ) public pure returns (uint) {
+        bool aIsFirst = a.created <= b.created;
```
sporejack, line 126: ...ternary?

sporejack (approved):

CalabashSquash (approved):
:madagascar:

contracts/lib/LibPerpetuals.sol
raymogg: @sporejack ternary seems a bit overkill here no? A ternary where you go `isThingTrue ? true : false` is the same as returning `isThingTrue`

kumar-ish (approved):
:fish:

PR #135, opened 2021-06-08
Verify on Deploy

raymogg:
# Motivation
In order to make debugging deployed versions of the protocol easier, it makes sense to have all the contract verified on Etherscan.

# Changes
- Changes FullDeploy to FullDeployTest - as its only used with test suites.
- Creates LiveDeploy for deploying to live networks
- Creates LiveVerify to be called after LiveDeploy (only if needed) to verify any contracts that failed to verify as part of the deploy.
- Removes DeployAndFund as it was redundant
- Update readme


README.md
```diff
@@ -26,41 +26,32 @@ yarn test
...
 To run individual test files, use
 ```
-npx hardhat test <PATH_TO_TEST>
+yarn hardhat test <PATH_TO_TEST>
 ```
 ## Deployment
 Hardhat does not support native deployments. To aid in deployment, hardhat-deploy is being used. You can see the deployment scripts in the `/deploy` directory.
 
-To run deploys
+Tracer also supports programatic verification of contracts during deployment. Use the `hardhat.config.js` file to set your Etherscan API key in order for this to be enabled.
```
sporejack:
```suggestion
Tracer also supports programmatic verification of contracts during deployment. Use the `hardhat.config.js` file to set your Etherscan API key in order for this to be enabled.
```

README.md
```diff
@@ -26,41 +26,32 @@ yarn test
...
+
+To run deploys to a live network
+```
+yarn hardhat deploy --network NETWORK --tags LiveDeploy
 ```
-npx hardhat deploy
+This will run the LiveDeployment script which is made to deploy to a real network. If there is any issues with verification, you may run
```
sporejack:
```suggestion
This will run the `LiveDeployment` script which is made to deploy to a real network. If there are any issues with verification, you may run
```

sporejack (changes requested):

deploy/LiveDeploy.js
```diff
@@ -0,0 +1,330 @@
...
+    let insurance = await tracerInstance.insuranceContract()
+    let pricing = await tracerInstance.pricingContract()
+    let liquidation = await tracerInstance.liquidationContract()
+
+    // Set Trader.sol to be whitelisted, as well as deployer (for testing purposes)
+    await tracerInstance.setWhitelist(trader.address, true)
+    await tracerInstance.setWhitelist(deployer, true)
+
+    // verify
+    await hre.run("verify:verify", {
```
CalabashSquash, line 242: might be a silly question, but why don't we just run all of these without `await`, since we don't actually do anything with the results, or need them to be finished before proceeding?

CalabashSquash (changes requested):

deploy/LiveDeploy.js
sporejack: Won't we still need to drive each promise?

raymogg: Yeah good point, I guess I just had the await to see if any failed and run them sequentually. Might hit rate limits on Etherscan? But i think that should be fine.

sporejack (approved):

deploy/LiveDeploy.js
raymogg: @sporejack @BenjaminPatch had a play around locally, it seems to make sense to just keep these async so its clear when failures occur. It doesn't add too much time to the flow either way. If it gets too annoying we can do them synchronously later on i feel

PR #137, opened 2021-06-09
update contract verification to avoid ambigious bytecode

raymogg:
# Motivation
Hardhat deploy had some trouble verifying certain contracts and needed explicit params to identify the contracts.

# Changes
- adds `contract` field to some deploys


kumar-ish (approved):
💯

sporejack (changes requested):
Linting is failing.

PR #138, opened 2021-06-09
Partial liquidation

CalabashSquash:
### Motivation
partial liquidation needs to be supported.

This means enforcing a minimum leftover amount (10x liquidation gas cost), and escrow amount is proportional to the amount being liquidated (escrow less if escrowing not the whole position).

### Changes
- Modify `LibLiquidation.calcEscrowLiquidationAmount(...)` to take amount being liquidated and total base amount
    - This is so the portion can be taken and multiplied by the whole escrow amount
- Add `LibLiquidation.partialLiquidationIsValid(...)`
    - Makes sure amount leftover > 10*usd cost of liquidation
- Fixed a bug in `LibLiquidation.liquidationBalanceChanges(...)` Where it was assumed `amount` was given with the same sign as the liquidated account's base.

**Note: ** functional liquidation tests are coming in another PR.


contracts/lib/LibLiquidation.sol
```diff
@@ -138,4 +152,23 @@ library LibLiquidation {
             return amountToReturn;
         }
     }
+
+    function partialLiquidationIsValid(
```
kumar-ish, line 175: Add documentation/NatSpec here

contracts/test/LibLiquidationMock.sol
```diff
@@ -72,4 +76,29 @@ library LibLiquidationMock {
             minimalReceipt
         );
     }
+
+    function partialLiquidationIsValid(
```
kumar-ish, line 83: Add documentation/NatSpec here

kumar-ish: Also, knowing which of these are WAD values would be helpful

contracts/lib/LibLiquidation.sol
```diff
@@ -28,14 +29,20 @@ library LibLiquidation {
 
     function calcEscrowLiquidationAmount(
```
kumar-ish, line 39: Add documentation/NatSpec here

kumar-ish (changes requested):

contracts/lib/LibLiquidation.sol
CalabashSquash: done

contracts/test/LibLiquidationMock.sol
CalabashSquash: probably redundant since mock but done

contracts/lib/LibLiquidation.sol
CalabashSquash: done

test/unit/LibLiquidation.js
```diff
@@ -334,4 +396,153 @@ describe("Unit tests: LibLiquidation.sol", function () {
...
+                const liquidatedBaseChange = ethers.utils.parseEther("-1")
+                const liquidatedQuoteChange = ethers.utils.parseEther("3")
+                const liquidatedBase = ethers.utils.parseEther("-145")
+                const liquidatedQuote = ethers.utils.parseEther("146")
+                const lastUpdatedGasCost = ethers.utils.parseEther("0.0003")
+                const liquidationGasCost = ethers.utils.parseEther("63516")
+                const price = ethers.utils.parseEther("1")
+
+                const valid = await libLiquidation.partialLiquidationIsValid(
+                    liquidatedBaseChange, // 1
```
kumar-ish: These numbers need to be changed (I think they were copied but not updated from the other test)

test/unit/LibLiquidation.js
```diff
@@ -334,4 +396,153 @@ describe("Unit tests: LibLiquidation.sol", function () {
...
+                    liquidatedQuoteChange, // 0
+                    liquidatedBase, // -145
+                    liquidatedQuote, // 0
+                    lastUpdatedGasCost, // 100
+                    liquidationGasCost, // 0
+                    price // 1
```
CalabashSquash:
```suggestion
                    liquidatedBaseChange,
                    liquidatedQuoteChange,
                    liquidatedBase,
                    liquidatedQuote,
                    lastUpdatedGasCost,
                    liquidationGasCost,
                    price
```

kumar-ish (approved):
🌞

contracts/lib/LibLiquidation.sol
```diff
@@ -138,4 +161,32 @@ library LibLiquidation {
...
+        int256 liquidatedBaseChange,
+        int256 liquidatedQuoteChange,
+        Balances.Account memory balanceToBeLiquidated,
+        uint256 liquidationGasCost,
+        uint256 price
+    ) internal pure returns (bool) {
+        uint256 minimumLeftoverMargin = PRBMathUD60x18.mul(
+            balanceToBeLiquidated.lastUpdatedGasPrice,
+            liquidationGasCost
+        ) * 10;
```
raymogg: Whats this magic number 10 here?

contracts/lib/LibLiquidation.sol
```diff
@@ -138,4 +161,32 @@ library LibLiquidation {
...
+        Balances.Position memory updatedPosition = Balances.Position(
+            balanceToBeLiquidated.position.quote + liquidatedQuoteChange,
+            balanceToBeLiquidated.position.base + liquidatedBaseChange
+        );
+        int256 margin = Balances.margin(updatedPosition, price);
+        return margin >= minimumLeftoverMargin.toInt256() || margin == 0;
```
raymogg: Bit worried about this `margin == 0` check. Technically margin as 0 could be a valid partial liquidation point. Should we also check the user has positions or something rather than just `margin == 0`. Also worth seeing if you can even get to this check with zero positions (base)

raymogg (approved):

contracts/lib/LibLiquidation.sol
CalabashSquash: Yeah true this is a magic number. Do you think this should be a market parameter? This is just what Garner and webby agreed on.

contracts/lib/LibLiquidation.sol
CalabashSquash: Yeah I think this should actually be return `margin >= minimumLeftoverMargin.toInt256() || updatedPosition.base == 0;`

Is that what you're saying?

contracts/Liquidation.sol
```diff
@@ -221,11 +247,20 @@ contract Liquidation is ILiquidation, Ownable {
             int256 liquidateeQuoteChange,
             int256 liquidateeBaseChange
         ) = LibLiquidation.liquidationBalanceChanges(
```
sporejack, line 248: Considering this snippet, this should really take a `Balances.Position`, no?

contracts/lib/LibLiquidation.sol
raymogg: So minimum leftover margin must be greater than 10 * liquidation gas cost, and min margin is some value + 6 * liquidation gas cost so this may or may not leave them above min margin?

Yeah I think set this as a constant for now. We could even make it configurable but lets keep it simple for now as a const.

contracts/Liquidation.sol
```diff
@@ -196,6 +198,30 @@ contract Liquidation is ILiquidation, Ownable {
         return amountToEscrow;
     }
 
+    /**
+     * @return true if the margin is greater than 10x liquidation gas cost (in quote tokens)
+     * @param liquidatedBaseChange How much base token to be liquidated
+     * @param liquidatedQuoteChange How much quote token to be liquidated
+     * @param balanceToBeLiquidated The balance of account to be liquidated
+     */
+    function checkPartialLiquidation(
```
sporejack, line 207: Should accept a `Balances.Position`.

contracts/lib/LibLiquidation.sol
sporejack: Even if it's hardcoded, it needs to be a defined constant somewhere.

contracts/test/LibLiquidationMock.sol
```diff
@@ -72,4 +76,32 @@ library LibLiquidationMock {
             minimalReceipt
         );
     }
+
+    /**
+     * @notice call LibLiquidation.partialLiquidationIsValid
+     */
+    function partialLiquidationIsValid(
```
sporejack, line 83: Extremely good candidate for `Balances.Position`.

sporejack: _Somewhat_ hacky in that it's not strictly a delta type, but you can still use this type for the _change in_ parameters, too.

sporejack (changes requested):
The API could be tidied up significantly by leaning into the use of `LibBalances` more.

contracts/test/LibLiquidationMock.sol
CalabashSquash: I've so far been unable to get this to work when calling external library functions (i.e. in these mock contracts). It always just throws with a some library-related error that is impossible to decypher and only works if you pass in individual values and wrap it in the struct inside the contract function.

contracts/Liquidation.sol
CalabashSquash: Yep good catches, I've changed it to `Balances.Position`. I could have chucked the updated position in the `Balances.Account` param but decided not to because that'll get confusing pretty quickly since that uses the `lastUpdatedGasPrice` of the *old* account balance so don't want to overlap the two

contracts/lib/LibLiquidation.sol
CalabashSquash: Actually, it will need to be `margin >= minimumLeftoverMargin.toInt256() || (updatedPosition.base == 0 && updatedPosition.quote == 0);` since if base == 0, quote can be negative and this position is still liquidatable.

sporejack (approved):

contracts/test/LibLiquidationMock.sol
```diff
@@ -81,27 +81,32 @@ library LibLiquidationMock {
...
+        uint256 price,
+        uint256 minimumLeftoverGasCostMultiplier
     ) external pure returns (bool) {
         Balances.Account memory balanceToBeLiquidated = Balances.Account(
-            Balances.Position(liquidatedQuote, liquidatedBase),
+            Balances.Position(0, 0), // Not used
             0,
             0,
             lastUpdatedGasPrice
         );
```
raymogg: Is a whole struct needed here if we only use the `lastUpdatedGasPrice`?

raymogg (approved):

PR #140, opened 2021-06-10
Fast Gas Price Oracle for Liquidation.sol

kumar-ish:
### Motivation 

There was previously a mismatch in the `TracerPerpetualSwaps.sol` contract with regards to what the gas oracle returned (a base token per gas price versus a gwei per gas price). This has now been fixed, and a fast gas oracle is now passed as a parameter to `Liquidation.sol` so that it can get the fast price of gas (instead of a $ amount previously)

### Changes

- ~Add fast gas oracle~ Start using fast gas oracle for liquidation, and normal quoteToken/Gas for TracerPerpetualSwaps.sol
- Make gas oracle uses consistent
- Deployment changes
- get rid of `gasPriceOracleAdapter` because it is just and adapter of an adapter
- verify the correct oracles on deployment


contracts/oracle/FastGasOracle.sol
```diff
@@ -0,0 +1,41 @@
...
+import "../Interfaces/IChainlinkOracle.sol";
+import "../lib/LibMath.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "prb-math/contracts/PRBMathUD60x18.sol";
+
+/**
+ * @dev The following is a sample Gas Price Oracle Implementation for a Liquidation Oracle.
+ *      It references the Chainlink fast gas price and returns the fast price of gas in wei.
+ */
+contract FastGasOracle is IOracle, Ownable {
```
CalabashSquash: You don't need to make a new contract for this. You can just use `contracts/oracle/Oracle.sol`, which is already done here: https://github.com/tracer-protocol/tracer-protocol/blob/contracts-v2/deploy/FullDeployTest.js#L55

deploy/FullDeployTest.js
```diff
@@ -125,6 +125,24 @@ module.exports = async function (hre) {
...
+        from: deployer,
+        log: true,
+        args: [fastGasOracle.address],
+        contract: "OracleAdapter",
+    })
+
+    console.log(
+        `Fast Gas Oracle Adapter Deployed ${gasPriceOracleAdapter.address}`
+    )
+
```
CalabashSquash: As per the above comment, https://github.com/tracer-protocol/tracer-protocol/blob/contracts-v2/deploy/FullDeployTest.js#L55 already deploys a gas oracle

CalabashSquash (changes requested):

deploy/FullDeployTest.js
```diff
@@ -268,6 +268,7 @@ module.exports = async function (hre) {
         "deployTracer",
         deployTracerData,
         oracleAdapter.address,
+        gasPriceOracle.address,
```
CalabashSquash:
```suggestion
        gasOracle.address,
```

Since `gasPriceOracle` is the quoteToken/gas oracle.

CalabashSquash (changes requested):

kumar-ish:
CI isn't running on this since I goofed it and am merging from the branch on my repo. I think it would be better to merge this a non-(contracts-v2) branch in tracer-protocol/tracer-protocol, then open another PR from that to contracts-v2

kumar-ish:
Redone with #147

PR #141, opened 2021-06-10
Execution price/amount mismatch fix

kumar-ish:
### Motivation

There was a bug where the price of execution of a trade was not accurate; this was caused by an erroneous `MatchedOrders` emission. This PR fixes that, along with an erroneous `FailedOrders` emission (whereby the right Long user was not being emitted), and fixes the correct amount of execution

### Changes
- Fix `MatchedOrders` and `FailedOrders` emissions in `TracerPerpetualSwaps` 
- Fix `recordPrice` to have correct amount of execution


raymogg (approved):

CalabashSquash (approved):
:macedonia:

PR #142, opened 2021-06-11
Insurance pool's quote in Tracer market can be 0

CalabashSquash:
### Motivation
It is OK for the insurance pool's quote to reach 0 inside its balance stored in `TracerPerpetualSwaps.sol`. However, we currently do a > (greater than) check instead of greater than or equal to.

### Changes
- Changed `require(balances[insuranceAddr].position.quote > 0)` to be >=
- Added 2 natspec comments


sporejack (approved):
LGTM!

PR #144, opened 2021-06-13
Liquidation functional tests

CalabashSquash:
### Motivation
we want Liquidation tests

### Changes
- Removed onlyTracer from `claimEscrow()`
- some natspec comments
- Removed duplicate `tx.gasPrice` check during `liquidate(...)`
- Set `escrowClaimed` to be true on `claimReceipt`
- Since `gasPriceOracleAdapter` in deployment was an adapter around an adapter, I deleted it. `GasOracle.sol` already makes sure it is in the right units.
- tests


contracts/Liquidation.sol
```diff
@@ -152,11 +159,6 @@ contract Liquidation is ILiquidation, Ownable {
         address account
     ) internal returns (uint256) {
         require(amount > 0, "LIQ: Liquidation amount <= 0");
-        // Limits the gas use when liquidating
-        require(
-            tx.gasprice <= IOracle(tracer.gasPriceOracle()).latestAnswer(),
-            "LIQ: GasPrice > FGasPrice"
-        );
```
raymogg: So the assumption is we now check this logic in the Tracer itself?

raymogg (approved):

sporejack (approved):

contracts/Liquidation.sol
CalabashSquash: I've changed this to now check in Liquidation.sol **instead** of TracerPerpsswaps.sol. Since fastGasOracle is in Liquidation.sol

raymogg (approved):

PR #145, opened 2021-06-14
Settle Functional Tests

raymogg:
# Motivation
Functional tests were needed for settlement. Also includes a fix for ensuring accounts are settled on deposit and withdraw, as well as optimisations in settle to reduce gas costs, and syncing of accounts that have no positions but have an index less than the current funding index.

# Changes
- settle on deposit and withdraw
- check for 0 base and sync account index
- settlement functional tests


CalabashSquash (approved):
:guinea_bissau: 

Can you give a quick run-down of why you can't go off the _current_ funding index?

raymogg:
> 🇬🇼
> 
> Can you give a quick run-down of why you can't go off the _current_ funding index?

Yep. The reason is if you use the currentFundingIndex, thats actually still being built (as in the pricing for the current hour isn't established yet) so you are essentially setting the user to be on a funding index that is not yet established.

sporejack (approved):

contracts/TracerPerpetualSwaps.sol
```diff
@@ -487,31 +491,35 @@ contract TracerPerpetualSwaps is
...
+        } else if (accountLastUpdatedIndex + 1 < currentGlobalFundingIndex) {
+            // Only settle account if its last updated index was before the last established
+            // global index this is since we reference the last global index
+            // Get current and global funding statuses
+            uint256 lastEstablishedIndex = currentGlobalFundingIndex - 1;
+            // Note: global rates reference the last fully established rate (hence the -1), and not
+            // the current global rate. User rates reference the last saved user rate
             Prices.FundingRateInstant memory currGlobalRate = pricingContract
-            .getFundingRate(pricingContract.currentFundingIndex() - 1);
+            .getFundingRate(lastEstablishedIndex);
```
CalabashSquash: What if you are settling() on `currentGlobalFundingIndex == 0`? Then `accountLastUpdatedIndex == 0`, so neither of these branches will be entered.

Also, does this mean there will be a deadlock, since on `deposit(...)`, when a user will have base == 0, the pricing contract's global funding index doesn't get updated (I don't think), so I don't see how a user could enter a branch that allows pricing to update it's funding index if you get what I mean.

I think I'm probably misunderstanding some of the code here but just wanted to make sure :bhutan:

CalabashSquash (changes requested):

contracts/TracerPerpetualSwaps.sol
raymogg: @BenjaminPatch yep when both `currentGlobalFundingIndex == 0` and `accountLastUpdatedIndex == 0` that is expected behaviour, settle should not do anything unless your index is lagging behind the global funding index.

In terms of deadlocking, the global index only gets updated on trades, not on settle calls itself. So you are right in saying anyone that deposits within the first hour won't update the global index, but as soon as someone executed a trade in the second hour, the global index will get updated.

Most of this logic occurs in `pricing.sol` and is triggered by the `pricing.recordTrade()` function

CalabashSquash (approved):
:sri_lanka:

PR #147, opened 2021-06-15
Gas price oracle

CalabashSquash:
### Motivation
There was previously a mismatch in the TracerPerpetualSwaps.sol contract with regards to what the gas oracle returned (a base token per gas price versus a gwei per gas price). This has now been fixed, and a fast gas oracle is now passed as a parameter to Liquidation.sol so that it can get the fast price of gas (instead of a $ amount previously)

### Changes
- ~Add fast gas oracle~ Start using fast gas oracle for liquidation, and normal quoteToken/Gas for TracerPerpetualSwaps.sol
- Make gas oracle uses consistent
- Deployment changes
- get rid of gasPriceOracleAdapter because it is just an adapter of an adapter
- verify the correct oracles on deployment


kumar-ish (approved):

deploy/LiveDeploy.js
```diff
@@ -102,13 +102,6 @@ module.exports = async function (hre) {
         contract: "GasOracle",
     })
 
-    const gasPriceOracleAdapter = await deploy("GasPriceOracleAdapter", {
```
raymogg, line 105: Just checking, we don't need this for the gas oracle? I believe that txn.gasPrice is reported in wei and the chainlink oracle reports gwei so we may need to convert units

raymogg (approved):

deploy/LiveDeploy.js
raymogg: Disregard. Saw your comment about it being a wrapper on a wrapper and that is correct

PR #149, opened 2021-06-17
Order execution price

CalabashSquash:
### Motivation
Since the `price` of an order, as stored in Trader.sol, is the price that the order was submitted with, rather than the actual execution price, a liquidator was able to sell a position with price == 0, where it emulates a market order. It would not get filled at price == 0, due to how order books work, but it would be stored with price == 0 nonetheless.

The calculation of slippage uses what is stored in Trader.sol to calculate the average price, btu this is not the execution price. It should instead use the actual average execution price of that order.

### Changes
- Add field `mapping(bytes32 => uint256) public averageExecutionPrice;` to Trader.sol
- Add function `calculateAverageExecutionPrice(...)` to LibPerpetuals.sol
    - This function takes the old rolling average, along with the old amount, and new amount + new average, and calculates the new average
    - e.g. if old average was $1 with 100 units, and the next 100 units are $0.5, the new average will be $0.75
- Use the average execution price of an order instead of just `order.price` on slippage calculations in `Liquidation.sol`


raymogg (approved):

contracts/lib/LibPerpetuals.sol
```diff
@@ -28,6 +28,29 @@ library Perpetuals {
...
+        uint256 oldFactor = PRBMathUD60x18.mul(oldFilledAmount, oldAverage);
+        uint256 newFactor = PRBMathUD60x18.mul(fillChange, newFillExecutionPrice);
+        uint256 newTotalAmount = oldFilledAmount + fillChange;
+        if (newTotalAmount == 0) {
+            return 0;
+        }
+        uint256 average = PRBMathUD60x18.div(oldFactor + newFactor, newTotalAmount);
+        return average;
+    }
+
```
sporejack, line 53: Should this be in `LibPerpetuals` or our internal `LibMath`? Realistically all it does is updates a running mean (as the doc comment states).

sporejack (commented):
Just an observation -- these changes can likely go through.

contracts/lib/LibPerpetuals.sol
CalabashSquash: I personally feel like it's more suited to `LibPerpetuals` just because it's a bit more context-focused. As in, while it is a pretty mathsy function, it's use-case in our situation is specific to order execution if you get what I'm saying. But I'm not too fussed, if you disagree I'll chuck it in LibMath.

PR #150, opened 2021-06-17
Make vars public

dospore:
### Motivation
- in order to calculate when the next funding rate is I need to know when the last funding rate was updated
- since both currentHour and startLast24Hours start at index  0 you can calculate time till next update using currentHour

### Changes
- make startLastHour and startLast24Hours public


CalabashSquash (approved):
lgtm :angola:

raymogg (approved):

PR #152, opened 2021-06-18
Add additional order matching predicates

kumar-ish:
## Motivation

We don't want someone to be able to make an order with themself, or orders to be matched in different markets. While the OME/other whitelisted trader would prevent this in any case, those extra checks have been added to `matchOrders` as a fail-safe.

## Changes

- Add predicates to make `matchOrders` check for the order makers being different, and the markets being the same for the two orders being matched
- Update tests

### DIFF
--- a/contracts/lib/LibPerpetuals.sol
+++ b/contracts/lib/LibPerpetuals.sol
@@ -1,6 +1,7 @@
 //SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity ^0.8.0;
 import "prb-math/contracts/PRBMathUD60x18.sol";
+import "hardhat/console.sol";
 
 library Perpetuals {
     enum Side {
@@ -96,7 +97,7 @@ library Perpetuals {
     /**
      * @notice Checks if two orders can be matched given their price, side of trade
      *  (two longs can't can't trade with one another, etc.), expiry times, fill amounts,
-     *  and time validation.
+     *  markets being the same, makers being different, and time validation.
      * @param a The first order
      * @param aFilled Amount of the first order that has already been filled
      * @param b The second order
@@ -114,11 +115,14 @@ library Perpetuals {
         bool opposingSides = a.side != b.side;
         // long order must have a price >= short order
         bool pricesMatch = a.side == Side.Long ? a.price >= b.price : a.price <= b.price;
+        bool marketsMatch = a.market == b.market;
+        bool makersDifferent = a.maker != b.maker;
         bool notExpired = currentTime < a.expires && currentTime < b.expires;
         bool notFilled = aFilled < a.amount && bFilled < b.amount;
         bool createdBefore = currentTime >= a.created && currentTime >= b.created;
 
-        return pricesMatch && opposingSides && notExpired && notFilled && createdBefore;
+        return
+            pricesMatch && makersDifferent && marketsMatch && opposingSides && notExpired && notFilled && createdBefore;
     }
 
     function getExecutionPrice(Order calldata a, Order calldata b) public pure returns (uint256) {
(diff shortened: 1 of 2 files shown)


contracts/lib/LibPerpetuals.sol
```diff
@@ -1,6 +1,7 @@
 //SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity ^0.8.0;
 import "prb-math/contracts/PRBMathUD60x18.sol";
+import "hardhat/console.sol";
```
CalabashSquash: This should be removed

contracts/lib/LibPerpetuals.sol
```diff
@@ -114,11 +115,14 @@ library Perpetuals {
         bool opposingSides = a.side != b.side;
         // long order must have a price >= short order
         bool pricesMatch = a.side == Side.Long ? a.price >= b.price : a.price <= b.price;
+        bool marketsMatch = a.market == b.market;
+        bool makersDifferent = a.maker != b.maker;
```
CalabashSquash, line 118: I'm wondering if this is important? If a user makes an order that gets successfully matched against another order made by themself, through the order book, wouldn't expected behaviour be to match them?
@sporejack wonder what you think about this.

CalabashSquash (changes requested):

contracts/lib/LibPerpetuals.sol
raymogg: @BenjaminPatch we discussed on Friday, I would say this is just a sanity check to avoid an manipulation on low liquidity books. Obviously the easiest way around this is just have two different accounts make the order.

If we treat the contracts as the absolute truth, it makes sense for them to check this predicate imo

raymogg (approved):

contracts/lib/LibPerpetuals.sol
CalabashSquash: Yeah but as you said it's trivial to circumvent it by using two different accounts

contracts/lib/LibPerpetuals.sol
kumar-ish: Oop

kumar-ish:
```suggestion
```

contracts/lib/LibPerpetuals.sol
raymogg: Yeah good point, I'll get Jacks opinion on this and we can make a decision. An extra check that isn't that valuable is kinda pointless.

contracts/lib/LibPerpetuals.sol
sporejack: https://hardhat.org/plugins/hardhat-log-remover.html

contracts/lib/LibPerpetuals.sol
sporejack: There were several defects found which would be mitigated by disallowing trading with oneself; however, given the refactor these are quite possibly mitigated entirely now. Provided `TCR-07` is fixed, wash trading shouldn't impact on-chain data in an adverse way. 

Conversely, what's the realistic downside to having this check in place? Aside from a UX perspective, gas is likely to be negligible (we're comparing two in-memory variables). Open to thoughts on this from both of you @raymogg and @BenjaminPatch .

CalabashSquash: If you think it's safer for this check, I'm fine with it. I guess we just have to make sure the OME also does this same check to make sure invalid order pairs aren't put on chain (I assume it already does)

Either way I'm easy

raymogg (approved):

sporejack (approved):

PR #155, opened 2021-06-22
Use appropriate filled amount in matchOrders

CalabashSquash:
### Motivation
TracerPerps.matchOrder had its own `filled` mapping, which it was reading from but never actually using (it was always zero). The Trader.sol contract was keeping track of filled amounts for each order, so matchOrder should use that.

### Changes
- Made matchOrders take `fillAmount` param, as it was being passed in by Trader.sol but not actually used
- Changed `TracerPerpetualSwaps.matchOrders(...)` to use the `Trader.filled` mapping to get filled amounts
    - Since this makes a contract function call now, tests can't call `matchOrders(...)` directly, so they have to go through a new TraderMock.sol
**TraderMock.sol**
- TraderMock.sol was added, which does `Trader.executeTrade(...)` without signature checks
- TraderMock.executeTrade is used by passing in a `Types.SignedLimitOrder` with the order inside, and the signature data just as dummy variables.
- I can't just pass in `Perpetuals.Order`s, because that violates the `ITrader.sol` interface.
---------------
- Go through the tests and use `TraderMock.executeTrade(...)` instead of `TracerPerpetualSwaps.matchOrders(...)`.
- Since now the orders' filled amounts actually get updated, you have to clear the filled amount if you specifically want to reuse an order
- An alternative is to make multiple of the same orders, just with slightly differing created times or something, so their order IDs will be different

### DIFF
--- a/contracts/Interfaces/IPricing.sol
+++ b/contracts/Interfaces/IPricing.sol
@@ -6,6 +6,8 @@ import "../lib/LibPrices.sol";
 interface IPricing {
     function getFundingRate(uint256 index) external view returns (Prices.FundingRateInstant memory);
 
+    function currentHour() external returns (uint8);
+
     function getInsuranceFundingRate(uint256 index) external view returns (Prices.FundingRateInstant memory);
 
     function currentFundingIndex() external view returns (uint256);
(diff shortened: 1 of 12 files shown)


sporejack (approved):
LGTM!

raymogg (approved):

PR #156, opened 2021-06-22
Add indexed keywords to events

kumar-ish:
## Motivation 

`TracerPerpetualSwaps.sol` had some events that should be indexed but weren't. This PR adds them.

## Changes

- Add indexed keyword to two events that should have them in


CalabashSquash (approved):
:fiji:

PR #157, opened 2021-06-22
Removed unused contracts

kumar-ish:
## Motivation

As a result of various changes, there were a few contracts that aren't used anymore. This PR removes them.

## Changes

- Remove `IDex.sol`, `Migrations.sol`, and `SafeMath96.sol`


CalabashSquash (approved):
:cameroon:

PR #159, opened 2021-06-22
Truncate messages in require statements

kumar-ish:
## Motivation

> Every reason string takes at least 32 bytes so make sure your string fits in 32 bytes or it will become more expensive

[Source](https://blog.polymath.network/solidity-tips-and-tricks-to-save-gas-and-reduce-bytecode-size-c44580b218e6)

There were a few reason strings in the contracts that had more than 32 characters. This pull request truncates them.

## Changes

- Change `require`'s in the various contracts to have a reason message that's <= 32 bytes

### DIFF
--- a/contracts/Insurance.sol
+++ b/contracts/Insurance.sol
@@ -185,7 +185,7 @@ contract Insurance is IInsurance, Ownable, SafetyWithdraw {
     }
 
     modifier onlyLiquidation() {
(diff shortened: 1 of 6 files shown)


kumar-ish:
Definitely did not forget to add tests to my original commit 👀

CalabashSquash (approved):

raymogg (approved):

PR #171, opened 2021-06-23
Address(0) checks

CalabashSquash:
check for address(0)


raymogg (approved):
If we are doing this, we should probably also check on inputs hey? Or use the `set` functions within constructors

PR #172, opened 2021-06-24
Address 0 check

CalabashSquash:
### Motivation and Changes
address(0) checks

### DIFF
--- a/contracts/TracerPerpetualSwaps.sol
+++ b/contracts/TracerPerpetualSwaps.sol
@@ -520,22 +520,27 @@ contract TracerPerpetualSwaps is ITracerPerpetualSwaps, Ownable, SafetyWithdraw 
     }
 
     function setLiquidationContract(address _liquidationContract) external override onlyOwner {
+        require(_liquidationContract != address(0), "address(0) given");
         liquidationContract = _liquidationContract;
     }
 
     function setInsuranceContract(address insurance) external override onlyOwner {
+        require(insurance != address(0), "address(0) given");
         insuranceContract = IInsurance(insurance);
     }
 
     function setPricingContract(address pricing) external override onlyOwner {
+        require(pricing != address(0), "address(0) given");
         pricingContract = IPricing(pricing);
     }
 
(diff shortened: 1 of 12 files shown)


contracts/TracerPerpetualSwaps.sol
```diff
@@ -569,11 +572,25 @@ contract TracerPerpetualSwaps is ITracerPerpetualSwaps, Ownable, SafetyWithdraw
         insurancePoolSwitchStage = _insurancePoolSwitchStage;
     }
 
+<<<<<<< HEAD
```
kumar-ish, line 575: Merge conflict here

contracts/deployers/PerpsDeployerV1.sol
```diff
@@ -25,6 +25,9 @@ contract PerpsDeployerV1 is IPerpsDeployer {
             _data,
             (bytes32, address, uint256, address, uint256, uint256, uint256, address, uint256, uint256, uint256)
         );
+        require(_tracerQuoteToken != address(0), "TCRDeploy: _tracerQuoteToken = address(0)");
```
kumar-ish: Over 32 bytes. If you remove the `address` it should be fine?

contracts/TracerPerpetualSwaps.sol
kumar-ish: Merge conflict here

kumar-ish (changes requested):

contracts/deployers/PerpsDeployerV1.sol
kumar-ish: Likewise with some of the ones below I think. Not a big deal but probably good practice to truncate them anyways. As an aside, we should talk to Dos about having codes/mappings for different reason messages; e.g. `TCRDeploy: ERR1` corresponds to `TCRDeploy: _tracerQuoteToken can't be the zero address`

raymogg (approved):

kumar-ish (approved):

PR #174, opened 2021-06-28
Remove margin check on settle

CalabashSquash:
### Motivation
People should not be prevented from settling their debts or anyone elses for any reason.

The check to see if below margin requirement should only be done on voluntary changes to a position.

### Changes
- Make `settle(...)` not check margin is valid.


kumar-ish (approved):
🕵️

PR #175, opened 2021-07-04
calcEscrowLiquidationAmount converts base to abs

CalabashSquash:
### Motivation
`LibLiquidation.calcEscrowLiquidationAmount(...)` takes base, but if base is negative, it multiplies escrow amount by a negative, which is less than 0, thus escrowed amount will be capped at 0 even though it shouldn't be (when liquidating a short).

### Changes
- Convert `base` to `base.abs()` before doing multiplication in `calcEscrowLiquidationAmount`

### DIFF
--- a/contracts/lib/LibLiquidation.sol
+++ b/contracts/lib/LibLiquidation.sol
@@ -9,7 +9,6 @@ import "prb-math/contracts/PRBMathSD59x18.sol";
 
 library LibLiquidation {
     using LibMath for uint256;
-    using LibMath for int256;
     using PRBMathUD60x18 for uint256;
     using PRBMathSD59x18 for int256;
 
@@ -34,7 +33,7 @@ library LibLiquidation {
      * @dev Assumes params are WAD
      * @param minMargin User's minimum margin
      * @param currentMargin User's current margin
-     * @param amount Amount being liquidated
+     * @param amount Absolute amount being liquidated
      * @param totalBase User's total base
      */
     function calcEscrowLiquidationAmount(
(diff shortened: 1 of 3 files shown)


raymogg (approved):

contracts/lib/LibLiquidation.sol
```diff
@@ -44,7 +43,11 @@ library LibLiquidation {
         int256 totalBase
     ) internal pure returns (uint256) {
         int256 amountToEscrow = currentMargin - (minMargin.toInt256() - currentMargin);
-        int256 amountToEscrowProportional = PRBMathSD59x18.mul(amountToEscrow, PRBMathSD59x18.div(amount, totalBase));
+        int256 absoluteBase = PRBMathSD59x18.abs(totalBase);
+        int256 amountToEscrowProportional = PRBMathSD59x18.mul(
+            amountToEscrow,
+            PRBMathSD59x18.div(amount, absoluteBase)
+        );
```
chrdevmar, line 50: When `amountToEscrow` is negative (`currentMargin` is less than half of the `minimumMargin`),  this would have previously resulted in a positive `amountToEscrowProportional` when `PRBMathSD59x18.div(amount, absoluteBase)` was negative. 

Since we are now forcing `PRBMathSD59x18.div(amount, absoluteBase)` to be positive, a negative `amountToEscrow` can never result in a positive `amountToEscrowProportional`.

Not sure if this would introduce an edge case for liquidating traders that are severely under margin?

CalabashSquash: If `amountToEscrow` is negative, `amountToEscrowProportional` should be negative too. That way we know when a liquidator should not be escrowing anything.

chrdevmar (approved):

PR #176, opened 2021-07-05
Add new insurance funding rate + add tests

kumar-ish:
# Motivation

Since the original whitepaper for Tracer's perpetual swaps was released, the way that the insurance funding rate is calculated has been changed. Below is the new formula:
image

Sorry for my chopped-up three second LaTeX :((

# Changes

- Update the way that the insurance funding rate is calculated and update corresponding unit tests

### DIFF
--- a/contracts/Insurance.sol
+++ b/contracts/Insurance.sol
@@ -200,18 +200,13 @@ contract Insurance is IInsurance {
     /**
      * @notice Gets the 8 hour funding rate for an insurance pool
      * @dev the funding rate is represented as
-     *      0.0036523 * (insurance_fund_target - insurance_fund_holdings) / leveraged_notional_value)
+     *      (182.648 / 8) * (5 * ((fundTarget - fundHoldings) / (fundTarget * 10_000))) ** 2
      */
     function getPoolFundingRate() external view override returns (uint256) {
-        // 0.0036523 as a WAD = 36523 * (10**11)
-        uint256 multiplyFactor = 36523 * (10**11);
-
-        uint256 levNotionalValue = tracer.leveragedNotionalValue();
-
-        // Traders only pay the insurance funding rate if the market has leverage
-        if (levNotionalValue == 0) {
-            return 0;
-        }
+        // Above equation can be refactored to have 0.000005707750000 as the constant in front
+        // (182.648 / 8) * (5 ** 2) * (1 / (10_000 ** 2)) = 0.000005707750000
+        // 0.000005707750000 as a WAD = 570775 * (10 ** 7)
+        uint256 multiplyFactor = 570775 * (10**7);
 
         uint256 poolHoldings = getPoolHoldings();
         uint256 poolTarget = getPoolTarget();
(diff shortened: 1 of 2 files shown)


test/unit/Insurance.js
```diff
@@ -629,26 +630,68 @@ describe("Unit tests: Insurance.sol", function () {
             })
         })
 
-        context("when the leveraged notional value is > 0", async () => {
-            it("returns the appropriate 8 hour funding rate", async () => {
+        context("when the pool is greater than the target", async () => {
```
CalabashSquash, line 633: Would you be able to write a test for when the pool is < the target && `(((fund_target - fund_holdings)/fund_target) ** 2) < 1`?

There are tests for the two edge cases (pool above target & pool has no holdings), but just for peace of mind a normal case test would be good.

CalabashSquash (changes requested):

contracts/Insurance.sol
```diff
@@ -221,7 +216,7 @@ contract Insurance is IInsurance {
             return 0;
         }
 
-        uint256 ratio = PRBMathUD60x18.div(poolTarget - poolHoldings, levNotionalValue);
+        uint256 ratio = PRBMathUD60x18.pow(PRBMathUD60x18.div(poolTarget - poolHoldings, poolTarget), 2 * 10**18);
```
CalabashSquash: Also, RBMath.pow(x, y) requires x > 1 in WAD (see https://github.com/hifi-finance/prb-math/blob/main/contracts/PRBMathUD60x18.sol#L357).

Since poolTarget - poolHoldings / poolTarget is always < 1, I'm not sure how this works. A test case as suggested in my other comment would check this.

CalabashSquash (changes requested):

contracts/Insurance.sol
```diff
@@ -200,18 +200,13 @@ contract Insurance is IInsurance {
     /**
      * @notice Gets the 8 hour funding rate for an insurance pool
      * @dev the funding rate is represented as
-     *      0.0036523 * (insurance_fund_target - insurance_fund_holdings) / leveraged_notional_value)
+     *      (182.648 / 8) * (5 * ((fundTarget - fundHoldings) / (fundTarget * 10_000))) ** 2
```
greenbookwebb, line 203: This could throw an error. div by 0/ Fund holdings could be positive

greenbookwebb (commented):
Looks good overall. I didn't go into the tests though. Let's just talk in person about tests if you want me to discuss it.

contracts/Insurance.sol
CalabashSquash: It could by this alone, but below there is a check for
```
        if (poolTarget <= poolHoldings) {
            return 0;
        }
```
if `poolTarget` is 0, then poolHoldings must be at least == pool target, and therefore will enter this branch and return 0

test/unit/Insurance.js
kumar-ish: Done

CalabashSquash (approved):

raymogg (approved):

PR #178, opened 2021-07-07
Timestamp on liquidate

CalabashSquash:
# Motivation
Determining the timestamp of a specific block locally can be quite annoying, the provider often doesn't know about a block for a while after it has happened.
This means a liquidator, when determining the creation time for their sell order, doesn't have a robust way of making sure the block.timestamp of receipt creation will be what they estimate off chain.

# Changes
- To solve for this, we can just emit the block's timestamp, used in receipt creation, so the liquidation bot can set `order.created` to this value.


kumar-ish (approved):
🏄

PR #182, opened 2021-07-08
Factor trade volume into hourly price instants

kumar-ish:
# Motivation

Currently, when calculating the average price in a given hour of a market, the cumulative price is set to be the sum of the prices of all the trades, ignoring the weight of the actual trades (i.e. the number of positions sold in that trade). This leads to asymmetric averages. 

e.g. Currently, for the following two trades that happen in an hour:
- 90 pos. @ $1.25
- 10 pos. @ $1.50

The average price for that hour is calculated to be (1.25 + 1.5) / 2 = $1.375, even though based on the actual positions that were filled, it should be (90x1.25 + 10x1.5) / 100 = $1.275

This PR solves this: https://github.com/code-423n4/2021-06-tracer-findings/issues/119

# Changes
- Factor trade volume into hourly price instants
- Misc. gas optimisation in `TracerPerpetualSwaps`

### DIFF
--- a/test/unit/LibPrices.js
+++ b/test/unit/LibPrices.js
@@ -3,35 +3,51 @@ const { ethers, getNamedAccounts, deployments } = require("hardhat")
 const { deploy } = deployments
 const zeroAddress = "0x0000000000000000000000000000000000000000"
 const MAX_INT256 = ethers.BigNumber.from(2).pow(255).sub(1)
-const calcExpectedTwaps = (oraclePrices, tracerPrices, hour) => {
+const calcExpectedTWAPs = (oraclePrices, tracerPrices, hour) => {
     let cumulativeDerivative = ethers.BigNumber.from("0")
     let cumulativeUnderlying = ethers.BigNumber.from("0")
-    let totalTimeWeight = ethers.BigNumber.from("0")
+    let totalDerivativeTimeWeight = ethers.BigNumber.from("0")
+    let totalUnderlyingTimeWeight = ethers.BigNumber.from("0")
 
     for (i = 0; i < 8; i++) {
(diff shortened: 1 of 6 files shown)


test/unit/LibPrices.js
```diff
@@ -317,12 +326,12 @@ describe("Unit tests: LibPrices.sol", function () {
...
                 for (i = 0; i < 24; i++) {
                     oraclePrices.push([
-                        ethers.utils.parseUnits((1 + 1 * i).toString(), 18),
-                        ethers.BigNumber.from("1"),
+                        ethers.utils.parseEther((1 + 1 * i).toString()),
+                        ethers.utils.parseEther("1"),
                     ])
                     tracerPrices.push([
-                        ethers.utils.parseUnits((1 + 0.5 * i).toString(), 18),
-                        ethers.BigNumber.from("1"),
```
CalabashSquash, line 325: Could you add a test that has non-uniform or non-1 `trades` values?
There's the highlighted one, which just has `trades == 1`, and one above, which puts both `price` and `trades` to equal `i + 1` on each. Would be good to have a test covering just like a typical case where both are sort of random

CalabashSquash (changes requested):

test/unit/LibPrices.js
kumar-ish: I'll add these in the `LibPricing` additional tests PR and resolve this once I have

kumar-ish: Nevermind, I did it in this PR; just have a look if that looks all good since it'd be good to get this merged in before the long awaited new functional tests

CalabashSquash (approved):

PR #184, opened 2021-07-09
SetPrice.js script, for generalised price setting

CalabashSquash:
# Motivation
It's handy having a bunch of set price scripts for Kovan. I started making a few, and there was quite a lot of duplicate code.

# Changes
- Make `scripts/SetPrice.js` which can be called from numerous places to reduce duplicate code


raymogg (commented):
Can scripts take in parameters? Would make this much nicer

raymogg (approved):

PR #185, opened 2021-07-12
Fix apply insurance and increase test coverage for LibPrices

kumar-ish:
# Motivation

Application of insurance funding rate was incorrectly calculating the difference between the users and system last updated funding rate, rather than the last updated cumulative funding rate.

There were some functions in LibPrices that didn't have test coverage for them. This pull request adds them.

# Changes
- Make the apply insurance function use the cumulative funding rate
- Add tests for `LibPrices.sol`
- Add functions being tested to mock contract
- Add some documentation to `LibPrices.sol`


test/unit/LibPrices.js
```diff
@@ -455,4 +482,237 @@ describe("Unit tests: LibPrices.sol", function () {
...
+                let userPosition = [
+                    ethers.utils.parseEther("0"),
+                    ethers.utils.parseEther("110"),
+                ] // quote, base
+                let insurancePosition = [
+                    ethers.utils.parseEther("0"),
+                    ethers.utils.parseEther("0"),
+                ] // quote, base
+                let insuranceGlobalRate = [0, ethers.utils.parseEther("10"), 0] // timestamp, fundingRate, cumulativeFundingRate
+                let insuranceUserRate = [0, ethers.utils.parseEther("5"), 0] // timestamp, fundingRate, cumulativeFundingRate
```
OsmanBran: @kumar-ish, @BenjaminPatch, @raymogg . The applyInsurance function does calculations based on the fundingRate not the cumulativeFundingRate. I think this is a major error as user's would be have differing insurance payments based on when they settle their accounts, it should be cumulativeFundingRate instead. Is this correct?

kumar-ish: @OsmanBran Yes, that's correct; fix accordingly

OsmanBran (dismissed):

CalabashSquash (approved):
Nice catch

PR #187, opened 2021-07-12
Check ERC20 transfer return value

OsmanBran:
# Motivation
The transfer and transferFrom functions from the ERC20 contract are used, however the return values of these functions are not checked.

# Changes
- Wrapped calls to these functions inside a require() to check if that the return value was true and the transfer was successful (code-423n4/2021-06-tracer-findings#115).


contracts/Insurance.sol
```diff
@@ -49,7 +49,7 @@ contract Insurance is IInsurance {
         // convert token amount to WAD
         uint256 quoteTokenDecimals = tracer.quoteTokenDecimals();
         uint256 rawTokenAmount = Balances.wadToToken(quoteTokenDecimals, amount);
-        collateralToken.transferFrom(msg.sender, address(this), rawTokenAmount);
+        require(collateralToken.transferFrom(msg.sender, address(this), rawTokenAmount), "transfer failed");
```
kumar-ish: It would be good to have a more informative error message; if you look at the other `require` statements in the contracts, they're (I think all) prefixed with a shorthand for the contract.

contracts/TracerPerpetualSwaps.sol
```diff
@@ -157,7 +157,7 @@ contract TracerPerpetualSwaps is ITracerPerpetualSwaps, Ownable, SafetyWithdraw
         // cast is safe since amount is a uint, and wadToToken can only
         // scale down the value
         uint256 rawTokenAmount = uint256(Balances.wadToToken(quoteTokenDecimals, amount).toInt256());
-        IERC20(tracerQuoteToken).transferFrom(msg.sender, address(this), rawTokenAmount);
+        require(IERC20(tracerQuoteToken).transferFrom(msg.sender, address(this), rawTokenAmount), "transfer failed");
```
kumar-ish: Likewise with the perp swaps contract

kumar-ish (changes requested):
As an aside, there's some gas savings with having `require` statements' messages be 32 bytes (characters) or less; see #159

CalabashSquash (approved):

CalabashSquash:
lint

kumar-ish (approved):
Lint && lgtm

PR #127, opened 2021-05-31
Liquidation calcUnitsSold tests

CalabashSquash:
### Motivation
Tests for liquidation need to be started

### Changes
- move `calcUnitsSold` to be above `calcAmountToReturn`
**Deployments**
- Change some minor deployment stuff
- Set decimals to 9 and gas price to 1 Gwei for gas oracle 
- Set decimals to 18 and price to 3000 USD for ETH oracle
- Fixed GetIntoLiquidatablePosition.js to make stable

- Added tests for calcUnitsSold in Liquidation.sol
    - More tests coming soon.

**Note:** Will also be moving the utility functions in `test/functional/Liquidation.js` to a new util file in a future PR.

PR #190, opened 2021-07-12
Fix fee subtraction logic issue

kumar-ish:
# Motivation
There was a logic error in fee subtraction from trades that slipped past code review during the lib refactor; the issue is that for long trades, the fee was being added to the quote of a user's position instead of subtracted. This pull request fixes that. 

Whether/how leveraged a trader is determined by the value of their absolute position minus their quote tokens. If the fee were to be added, then they would become less leveraged after the fee is applied (as that means their quote would increase but the notional value of their position would stay the same). In the same vein, having a fee that's subtracted from their quote makes them ever so slightly more leveraged. 

Fixes [this](https://github.com/code-423n4/2021-06-tracer-findings/issues/127).

# Changes
- Fix fee subtraction issue
- Corresponding changes to tests
- Some stylistic things in `LibBalances`
- Add coverage directory to `.prettierignore` (to avoid Hardhat-generated test coverage files from being linted)

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -8,3 +8,4 @@ types/*
 cache/*
 *.md
 *.tmp
+coverage/*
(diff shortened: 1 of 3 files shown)


CalabashSquash (approved):
:aruba:

PR #191, opened 2021-07-12
Offset funding rate calculation by a factor of 8

kumar-ish:
# Motivation

The funding rate is calculated on an eight-hourly basis, but is paid on an hourly basis. This PR offsets the funding rate calculation by a factor of eight to account for that. 

# Changes
- Add offset for funding rate calculation
- Change corresponding test

### DIFF
--- a/contracts/Pricing.sol
+++ b/contracts/Pricing.sol
@@ -39,6 +39,10 @@ contract Pricing is IPricing {
     uint256 public startLast24Hours;
     uint8 public override currentHour;
 
+    // The funding rate is supposed to be 8-hourly, but because it's paid hourly
+    // in the contracts, we offset the calculated rate by a factor of 8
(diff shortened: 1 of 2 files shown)


CalabashSquash (approved):

contracts/Pricing.sol
```diff
@@ -39,6 +39,10 @@ contract Pricing is IPricing {
     uint256 public startLast24Hours;
     uint8 public override currentHour;
 
+    // The funding rate is supposed to be 8-hourly, but because it's paid hourly
+    // in the contracts, we offset the calculated rate by a factor of 8
+    int256 internal constant FUNDING_RATE_OFFSET = 8;
```
kumar-ish:
```suggestion
    int256 private constant FUNDING_RATE_OFFSET = 8;
```

PR #192, opened 2021-07-12
Mark state variables immutable where possible

sporejack:
# Motivation
During [the Code Arena audit](https://github.com/code-423n4/2021-06-tracer-findings), it was flagged (specifically in [#40](https://github.com/code-423n4/2021-06-tracer-findings/issues/40)) that some state variables are likely able to be marked as `immutable`.

# Changes
 - Mark `collateralAsset` in `Insurance.sol` as `immutable`
 - Mark `token` in `Insurance.sol` as `immutable`
 - Mark `insuranceContract` in `Liquidation.sol` as `immutable`
 - Mark `fastGasOracle` in `Liquidation.sol` as `immutable`
 - Mark `tracer` in `Pricing.sol` as `immutable`

### DIFF
--- a/contracts/Liquidation.sol
+++ b/contracts/Liquidation.sol
@@ -27,8 +27,8 @@ contract Liquidation is ILiquidation, Ownable {
     uint256 public override minimumLeftoverGasCostMultiplier = 10;
     IPricing public pricing;
     ITracerPerpetualSwaps public tracer;
-    address public insuranceContract;
-    address public fastGasOracle;
+    address public immutable insuranceContract;
+    address public immutable fastGasOracle;
 
     // Receipt ID => LiquidationReceipt
     mapping(uint256 => LibLiquidation.LiquidationReceipt) public liquidationReceipts;
--- a/contracts/Insurance.sol
+++ b/contracts/Insurance.sol
@@ -17,10 +17,10 @@ contract Insurance is IInsurance {
     using LibMath for int256;
     ITracerPerpetualsFactory public perpsFactory;
 
-    address public collateralAsset; // Address of collateral asset
+    address public immutable collateralAsset; // Address of collateral asset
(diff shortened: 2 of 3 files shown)


raymogg (approved):

contracts/Liquidation.sol
```diff
@@ -27,8 +27,8 @@ contract Liquidation is ILiquidation, Ownable {
     uint256 public override minimumLeftoverGasCostMultiplier = 10;
     IPricing public pricing;
     ITracerPerpetualSwaps public tracer;
-    address public insuranceContract;
-    address public fastGasOracle;
+    address public immutable insuranceContract;
+    address public immutable fastGasOracle;
```
kumar-ish, line 31: This is actually my bad from the PR that I made for adding a fast gas oracle to `Liquidation`, but you should be able to change the address of this variable. Could you add a setter to it please? (and consequently remove the immutable keyword)

sporejack: @raymogg @BenjaminPatch Do we creep scope or would we prefer another PR? Particularly worried about correspondence with Code Area issues.

kumar-ish (changes requested):
The issue that you were referring to was done with manual analysis and not comprehensive; refer to [this](https://github.com/code-423n4/2021-06-tracer-findings/issues/24), done by someone with a custom Solidity compiler (note that some of the variables such as `insurance` in `Pricing.sol` weren't picked up by the one you mentioned).

contracts/Liquidation.sol
kumar-ish: That is actually true; probably good to leave it to another PR.

OsmanBran:
Closing as issue fixed by https://github.com/tracer-protocol/perpetual-contracts/pull/206

PR #193, opened 2021-07-12
Refactor constants

kumar-ish:
# Motivation 
There appear to be some marginal gas improvements to having constants in contracts instead of Solidity's inlined numbers (see [here](https://github.com/jmcph4/solidity-inlined-constants)). This PR refactors such variables in the contracts. This also makes it less magic number-y. 

EDIT: I was only going to factor out the constants in the insurance contract initially, but increased the scope to all the contracts, thus the bad branch name :c

# Changes
- Refactor numbers in the contracts as constants.


CalabashSquash (approved):
:timor_leste:

kumar-ish:
Closes https://github.com/code-423n4/2021-06-tracer-findings/issues/14 and https://github.com/code-423n4/2021-06-tracer-findings/issues/50

PR #194, opened 2021-07-13
Refactored liquidation contract

OsmanBran:
# Motivation
Note the branch is incorrectly named as claim-escrow but involves refactoring changes to the liquidation contract.

The claimEscrow() function can be changed to external for gas savings and clarity.

The claimEscrow() comment also states that only a trader can call the function, however anyone can call the function.

Gas savings in verifyAndSubmitLiquidation() due to minimumMargin variable being calculated twice.

# Changes
- claimEscrow() function changed from public to external (code-423n4/2021-06-tracer-findings#128).
- claimEscrow() comment updated (code-423n4/2021-06-tracer-findings#26).
- saved the minimumMargin variable to memory to reduce gas (code-423n4/2021-06-tracer-findings#131).


kumar-ish (approved):
:surfer:

PR #196, opened 2021-07-13
Remove insurance constructor event

OsmanBran:
# Motivation
There is inconsistency in the use of events for contract constructors. Insurance contains an event in its constructor, but no other contracts do (code-423n4/2021-06-tracer-findings#10).

# Changes
The insurance constructor event has been removed. All other contracts can be referenced from the Tracer, which has an event emitted from the factory.


kumar-ish (approved):
:surfer:

PR #198, opened 2021-07-14
Ensure GasOracle latestAnswer() is in WAD format

OsmanBran:
# Motivation
The latestAnswer() function assumes that oracle latest answer returns values in WAD format. This may not necessarily be true, an explicit check and conversion should be added (code-423n4/2021-06-tracer-findings#93).

# Changes
- Added two functions that check oracle answer decimals. Function will revert if max number of decimals is exceeded in answer (i.e number provided with more than 18 decimals). Otherwise number is scaled up to 18 decimals to convert to WAD.

### DIFF
--- a/contracts/oracle/GasOracle.sol
+++ b/contracts/oracle/GasOracle.sol
@@ -18,6 +18,7 @@ contract GasOracle is IOracle, Ownable {
     IChainlinkOracle public priceOracle;
     uint8 public override decimals = 18;
     uint256 private constant MAX_DECIMALS = 18;
+    uint256 private constant MAX_GWEI_DECIMALS = 9;
 
     constructor(address _priceOracle, address _gasOracle) {
         require(_gasOracle != address(0), "GasOracle: _gasOracle == 0");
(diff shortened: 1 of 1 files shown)


contracts/oracle/GasOracle.sol
```diff
@@ -31,26 +32,34 @@ contract GasOracle is IOracle, Ownable {
...
-        IChainlinkOracle oracle = IChainlinkOracle(_oracle);
-        // reset the scaler for consistency
-        uint8 _decimals = oracle.decimals(); // 9
+    function _gasPriceToWad() internal view returns (uint256) {
+        uint256 result = uint256(gasOracle.latestAnswer());
+        uint8 _decimals = gasOracle.decimals();
+        require(_decimals <= MAX_GWEI_DECIMALS, "GAS: too many decimals");
+        uint256 scaler = uint256(10**(MAX_GWEI_DECIMALS - _decimals));
+        return result * scaler;
+    }
```
CalabashSquash, line 52: Do we want an ability to take gas oracles that use higher precision than 9 decimal places?

e.g. a gas oracle that has 18 decimals, in which case we would want to divide by 10^9

OsmanBran: I was considering this, but since the oracle reports in GWEI, would it even be possible for an oracle to have a higher precision than 10^18?

CalabashSquash: Yeah I don't think so.

CalabashSquash (approved):

PR #199, opened 2021-07-14
Remove references to safety withdraw

raymogg:
# Motivation
Having a safety withdraw method that allows the owner to remove any token from the market isn't a great safety mechanism, as it enables rugs.

Since the only token moving in and out of the `TracerPerpetualSwap.sol` contract is the margin token that is immutable, this method has been removed completely. Other failsafes will instead be used.

Solves this issue: https://github.com/code-423n4/2021-06-tracer-findings/issues/81

# Changes
- remove SafetyWithdraw and all references of it


OsmanBran (approved):
LGTM

kumar-ish (approved):
:surfing_man:

PR #200, opened 2021-07-15
Remove calls to deprecated Aggregator API

OsmanBran:
# Motivation
The code base uses the function `latestAnswer` when using raw Chainlink price feeds. This is deprecated and `latestRoundData` should be used instead. Additionally, there is no implementation of an up-to-date Chainlink feed contract in the codebase, one should be created to account for the purpose of testing.

Any oracles used by Tracer must be approved by the community and conform to the IOracle interface. The IOracle interface contains a latestAnswer() function which must return values in WAD format. To achieve this:
- All Chainlink Oracle feeds (except for the Fast Gas / GWEI feed) must be wrapped in a Chainlink Adapter to ensure the IOracle interface is conformed to (WAD formatting).

# Changes
Removed references to deprecated Aggregator API (code-423n4/2021-06-tracer-findings#145)
- A mock implementation of a current Chainlink contract was created which implements the latestRoundData function. The sample Oracle contract was deleted.
- Any direct calls to Chainlink feeds with latestAnswer were replaced with latestRoundData 
- Deprecated functions removed from IChainlinkOracle interface
- Deployment scripts were updated in line with contract changes.


contracts/oracle/ChainlinkOracle.sol
```diff
@@ -0,0 +1,87 @@
...
+        revert("CO: Deprecated function");
+    }
+
+    function getTimestamp(uint256 roundId) external view override returns (uint256) {
+        revert("CO: Deprecated function");
+    }
+
+    function description() external view override returns (string memory) {
+        revert("CO: Deprecated function");
+    }
```
OsmanBran: @raymogg Should we remove these deprecated calls from the interface? The ChainlinkOracleAdapter no longer supports these.

raymogg: Yeah I think we may as well remove them @OsmanBran since we are staying away from the deprecated api

raymogg (changes requested):

contracts/oracle/ChainlinkOracle.sol
```diff
@@ -0,0 +1,87 @@
...
+            uint80 roundId,
+            int256 answer,
+            uint256 startedAt,
+            uint256 updatedAt,
+            uint80 answeredInRound
+        )
+    {
+        revert("CO: Deprecated function");
+    }
+
```
CalabashSquash: Why are we saying this is deprecated?

https://docs.chain.link/docs/price-feeds-api-reference/#getrounddata

OsmanBran: Yep, it's not, I'll update now.

OsmanBran: @BenjaminPatch the getRoundData function has been removed altogether from the IChainlinkOracle interface as we don't use it to fetch answers in the system

contracts/Interfaces/README_Interfaces_Overview.md
```diff
@@ -50,10 +50,9 @@ IInsurance.sol is an interface that describes the functions that an Insurance co
...
-**Explanation:** This interface described the minimum functionality that a Tracer/Gas Oracle need to have. Each oracle can have different implementations (although they must conform to IOracle.sol), the Oracle must be community approved. Chainlink reference data contracts currently conform to the IOracle specification (as of 04/11/2020) and as such Chainlink data contracts can be used as the oracle implementation. 
+IOracle.sol is an interface that describes the functions that an Oracle Feed contract should implement.   
+=>The specific implementation of this interface as a contract can be found in GasOracle.sol (An example an oracle that references Chainlink fast gas price and ETH/USD price to get a gas cost in $USD)
+**Explanation:** This interface described the minimum functionality that a Tracer/Gas Oracle need to have. Each oracle can have different implementations (although they must conform to IOracle.sol), the Oracle must be community approved. Chainlink feed contracts (excluding the Fast Gas / GWEI feed) must be wrapped in a Tracer Chainlink Adapter which conforms to the IOracle specification (ChainlinkOracleAdapter.sol).
```
CalabashSquash: Add an extra newline before **Explanation**, just because github renderer needs an extra space to newline it. (View file to see what I mean)

CalabashSquash (changes requested):

CalabashSquash (approved):

raymogg (approved):

deploy/KovanDeploy.js
```diff
@@ -88,12 +88,12 @@ module.exports = async function (hre) {
...
 
-    // takes in the chainlink oracle wrapped (to return 10^18) and a gas price oracle
-    // not wrapped (to return 10^9)
+    // takes in the Fast Gas/Gwei and ETH/USD Chainlink oracles (not wrapped in adapter)
+    // provides the USD/Gas price in WAD format
     const gasPriceOracle = await deploy("GasPriceOracle", {
         from: deployer,
         log: true,
-        args: [ethOracleAdapter.address, gasOracle.address],
+        args: [ethOracle, gasOracle.address],
```
raymogg, line 96: Double sanity check: Why are we no longer wrapping the eth price oracle here?

raymogg (changes requested):

deploy/KovanDeploy.js
OsmanBran: No longer wrapped because we assume that both oracles are given to the gasOracle as raw chainlink feeds (seen by the IChainlinkOracle interfaces used in gasOracle.sol). The gasOracle then ensures that it converts these raw chainlink answers to WAD format to conform to the IOracle standard.

Prior to this change the ethOracleAdapter was providing the gasOracle with WAD formatted answers, however it was incorrectly being passed through the IChainlinkOracle (which suggested it was a raw feed and potentially unscaled). Hence why it was called out during audit for potentially not being scaled.

raymogg (approved):

PR #207, opened 2021-07-20
Add checks for user set parameters

OsmanBran:
# Motivation
Code Arena audit revealed possible unintended behaviour from unchecked params.

# Changes
Added check to maintain leverage and pool switch relationship (code-423n4/2021-06-tracer-findings#117)
- Require that lowestMaxLeverage < maxLeverage in constructor and setters
- Require that insurancePoolSwitchStage < deleveragingCliff in constructor and setters
- Added relevant unit tests

Added check to require that percent values don't exceed 100% (code-423n4/2021-06-tracer-findings#102)
InsurancePoolSwitchStage was not checked since insurancePoolSwitchStage < deleveragingCliff.
- Require that setFeeRate <= 100% in constructor and setters
- Require that deleveragingCliff <= 100% in constructor and setters
- Added relevant unit tests

Added check for maxSlippage and emitted events for other critical setters (code-423n4/2021-06-tracer-findings#77)
- Require that maxSlippage <= 100% in constructor and setters + unit test
- Emit events for setMaxSlippage, setMinimumLeftoverGasCostMultiplier and setReleaseTime


CalabashSquash (dismissed):
Me likey

kumar-ish (changes requested):
`feeRate` is a WAD value as a decimal; not a percent unlike some of the insurance/liquidation-specific variables. Refer to the NatSpec constructor to see what's what; it should have a max of `10**18`

contracts/TracerPerpetualSwaps.sol
```diff
@@ -564,10 +570,12 @@ contract TracerPerpetualSwaps is ITracerPerpetualSwaps, Ownable {
     }
 
     function setFeeRate(uint256 _feeRate) external override onlyOwner {
+        require(_feeRate <= MAX_PERCENT, "TCR: Fee rate > 100%");
         feeRate = _feeRate;
     }
 
     function setMaxLeverage(uint256 _maxLeverage) external override onlyOwner {
```
kumar-ish, line 601: Should there also be a check for max leverage being required to be greater than (or equal to) 1 in WAD?

kumar-ish: Likewise with lowest max leverage

kumar-ish (changes requested):

contracts/TracerPerpetualSwaps.sol
CalabashSquash: It's not theoretically impossible for it to be less than 1.

contracts/TracerPerpetualSwaps.sol
```diff
@@ -33,6 +33,8 @@ contract TracerPerpetualSwaps is ITracerPerpetualSwaps, Ownable {
     uint256 public override fees;
     address public override feeReceiver;
 
+    uint256 private constant MAX_PERCENT = 1e20; // Constant for 100% in WAD format, used to check valid parameters
```
kumar-ish: Slightly nicer way of writing this would be `100e18` (conveys that it's 100 in WAD)

contracts/TracerPerpetualSwaps.sol
kumar-ish: When would you want that? That sounds like overcollateralised accounts

OsmanBran: Since they should theoretically hold (fee will be valid as long as it's not negative, same with leverage), should these just be left to our current trust model where the Owner decides and accepts the risk?

kumar-ish:
If you could also add parameter validation for `maxSlippage` in the liquidation contract, that'd be good too (and any other user set parameters in the contracts that I might be missing)

contracts/TracerPerpetualSwaps.sol
kumar-ish: I'm not sure tbh, see what Patchy's decision is

CalabashSquash: This PR isn't for checking what is desirable. it is for making sure Things that would actually break the market are not allowed. Having a market that only allows for overcollateralised positions would be weird, but it is still definitely possible and would work perfectly fine.

So I would say it should be fine to allow for <1x leverage

OsmanBran:
> If you could also add parameter validation for `maxSlippage` in the liquidation contract, that'd be good too (and any other user set parameters in the contracts that I might be missing)

Resolved in latest commit, along with other issues mentioned in C4 audit issue 77.

contracts/Liquidation.sol
```diff
@@ -45,6 +47,9 @@ contract Liquidation is ILiquidation, Ownable {
         uint256 timestamp
     );
     event InvalidClaimOrder(uint256 indexed receiptId);
+    event ReleaseTimeUpdated(address indexed owner, uint256 releaseTime);
+    event MinimumLeftoverGasCostMultiplierUpdated(address indexed owner, uint256 minimumLeftoverGasCostMultiplier);
+    event MaxSlippageUpdated(address indexed owner, uint256 maxSlippage);
```
kumar-ish: Is there any reason to have the addresses in these events be indexed?

OsmanBran: Left as indexed to allow for users to sort through events by the owner. However, since owner changes are probably not that frequent I can remove.

kumar-ish: These should all have tests associated with them; i.e. for the setters that emit these events, we should have a test checking they emit them (`expect(...).to.emit(...).withArgs(...)`)

kumar-ish: I guess it's more so that it's kind of independent of the owner (i.e. who the owner is doesn't really matter anyways). I think it's not worth having the owner param in the events anyways

kumar-ish:
The `transferOwnership` function should also have an event emitted

OsmanBran:
> The `transferOwnership` function should also have an event emitted

Ownership transfers are planned to go through DAO proposal which will be highly visible, so an event for the purpose of transparency isn't necessary imo

kumar-ish:
That's only for DAO-deployed markets; once permissionless markets are allowed, anyone could deploy a market and it doesn't need to go through a proposal.

kumar-ish (approved):
:nicaragua:

PR #212, opened 2021-07-22
Account for no trade periods in pricing

OsmanBran:
# Motivation
C4 audit revealed that the `averagePrice` and `averagePriceForPeriod` did not account for periods with no trades which could lead to division by 0 errors. This would cause the `recordTrade` function to revert on the first trade in some cases.

Additionally, the hourly price arrays were only updated in hours where trades occurred, which can lead to misleading 24 hour periods used in the `get24HourPrices` function.

# Changes
`averagePrice` function updated to handle periods with no trades (https://github.com/code-423n4/2021-06-tracer-findings/issues/139)
- Return `MAX_INTEGER` instead of 0 if no trades occurred
- Update `getHourlyAvgOraclePrice` and `getHourlyAvgTracerPrice` which use `averagePrice` descriptions
- Update `recordTrade` which uses `getHourlyAvgTracerPrice` to only update if trades occurred

`averagePriceForPeriod` updated to handle periods with no trades (https://github.com/code-423n4/2021-06-tracer-findings/issues/140)
 **This was previously causing an error as `averagePriceForPeriod` was called with a period with no trades when the first trade was sent 24 hours after deployment**
- Return `MAX_INTEGER` instead of reverting if no trades occurred
- Update `get24HourPrices` which uses `averagePriceForPeriod` to return MAX INTEGER
- Update `updateTimeValue` which uses `get24HourPrices` to only update if trades occur in period
- Unit tests

Update `recordTrade` to correctly dismiss stale pricing (https://github.com/code-423n4/2021-06-tracer-findings/issues/142)
- Update the `recordTrade` function to update all hours that have passed since the last trade
- Update the `updatePrice` function to take in an hour parameter so that a local variable can be used to increment time
- 24 hour periods now accurately reflect 24 hours of length for the `get24HoursPrices` function
- Functional tests


contracts/Pricing.sol
```diff
@@ -76,33 +76,42 @@ contract Pricing is IPricing {
...
+            // emit the hourly average and udpate funding rate if trades occurred
+            if (hourlyTracerPrice != type(uint256).max) {
+                // emit the old hourly average
+                emit HourlyPriceUpdated(hourlyTracerPrice, currentHour);
+
+                // update funding rate for the previous hour
+                updateFundingRate();
+            }
+
+            uint256 elapsedHours = (block.timestamp - startLastHour) / 3600;
```
kumar-ish, line 93: This should be capped at 24 (ternary works) -- for a market that hasn't been active for a few weeks/months (however unlikely), this would go into a very long loop

contracts/Pricing.sol
```diff
@@ -76,33 +76,42 @@ contract Pricing is IPricing {
...
+            // if more than one hour passed, update any skipped hour prices as 0
+            if (elapsedHours > 1) {
+                for (uint256 i = 0; i < elapsedHours - 1; i++) {
+                    currentHour = (currentHour + 1) % 24;
+                    updatePrice(0, 0, 0, true);
```
kumar-ish: I wonder if this is the most efficient way to do this -- this is making a change to a global state variable at every turn even though you know what it is going to be at the end. Maybe it be worth making `updatePrice` use a pure function in the if-bodies or something, and reuse that function here?

contracts/lib/LibPrices.sol
```diff
@@ -74,6 +73,11 @@ library Prices {
             }
         }
 
+        // return max integer if no trades occurred in the last 24 hours
+        if (j == 0) {
```
kumar-ish, line 77: Need to add a unit test for this

kumar-ish (dismissed):

contracts/Pricing.sol
OsmanBran: Updated function to take in an hour variable, then used a local variable `staleHour` in for loop to address this.

raymogg (approved):

CalabashSquash (approved):
:kiribati:

PR #215, opened 2021-07-23
Use external modifier where possible

OsmanBran:
# Motivation
Address https://github.com/code-423n4/2021-06-tracer-findings/issues/62
Note that only setMaxSlippage was changed. Other functions outlined cannot be set to external as they override public functions from OpenZeppelin.

Additional functions set to external after analysis with Slither

# Changes
- Updated setMaxSlippage
- Updated functions as per Slither analysis


kumar-ish (commented):
This looks good, but I'll have another look :soon::tm: to make sure everything is covered (from memory there was an edge case that Slither didn't pick up on)

kumar-ish (approved):
:kite:

PR #216, opened 2021-07-23
Make liquidation gas cost upgradeable

OsmanBran:
# Motivation
Address https://github.com/code-423n4/2021-06-tracer-findings/issues/101
Liquidation gas cost may change in the future so should not be hardcoded.

# Changes
- Added liquidation gas cost as a parameter in constructor rather than a constant
- Created a liquidation gas cost setter so owner can change in future


kumar-ish (approved):
:shield:

PR #217, opened 2021-07-23
Remove less than redundant check

OsmanBran:
# Motivation
https://github.com/code-423n4/2021-06-tracer-findings/issues/17

# Changes
- Change `if (amountToReturn <= 0)` to `if (amountToReturn == 0)`. `amoutToReturn` is a uint256 and will never be less than 0.


kumar-ish (approved):
:surfer:

PR #220, opened 2021-07-27
fix: WAD error in withdrawFees and dust removal

OsmanBran:
# Motivation
The test token used has 18 decimals, makes it impossible to test WAD conversions and dust removal mechanisms.

The withdrawFees does not convert fees value back to token representation which causes an error when ERC20.transfer is called (i.e more funds are requested than should be since WAD value has been scaled up).

There is inconsistency between the deposit and withdraw functions regarding dust in emitted events and TVL calculations

# Changes
Change the test token to have 8 decimals

In `withdrawFees`, convert `fees` from WAD representation back to token representation to prevent error thrown during transfer call.

`Deposited` event emits the amount deposited with dust, while the `Withdraw` event does not (https://github.com/code-423n4/2021-06-tracer-findings/issues/56)
- Emit the deposited amount with dust removed instead of the raw amount
- Amend Natspec comments to reflect events
- Unit tests

Withdraw TVL calculation updates including dust, whereas deposit does not (https://github.com/code-423n4/2021-06-tracer-findings/issues/57)
- Use value with dust removed for TVL calculation
- Unit test

### DIFF
--- a/contracts/TracerPerpetualSwaps.sol
+++ b/contracts/TracerPerpetualSwaps.sol
@@ -231,7 +231,7 @@ contract TracerPerpetualSwaps is ITracerPerpetualSwaps, Ownable {
         _updateAccountLeverage(msg.sender);
 
         // Safemath will throw if tvl < amount
-        tvl = tvl - amount;
+        tvl = tvl - uint256(convertedWadAmount);
 
         // perform transfer
         require(IERC20(tracerQuoteToken).transfer(msg.sender, rawTokenAmount), "TCR: Transfer failed");
--- a/contracts/TracerPerpetualSwaps.sol
+++ b/contracts/TracerPerpetualSwaps.sol
@@ -164,7 +164,8 @@ contract TracerPerpetualSwaps is ITracerPerpetualSwaps, Ownable {
 
     /**
      * @notice Allows a user to deposit into their margin account
-     * @dev this contract must be an approved spender of the markets quote token on behalf of the depositer.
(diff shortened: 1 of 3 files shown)


kumar-ish (changes requested):
I would say to write some tests where you deposit some amount with a dust amount (e.g. `5.000001` ethers) and check for the resulting event emission to have it removed (if I understand all this correctly). Likewise for any other similar tests that might require it.

contracts/TracerPerpetualSwaps.sol
```diff
@@ -553,9 +555,12 @@ contract TracerPerpetualSwaps is ITracerPerpetualSwaps, Ownable {
...
         tvl = tvl - tempFees;
 
+        // Convert fees from WAD format to token representation
+        uint256 rawTokenFees = Balances.wadToToken(quoteTokenDecimals, tempFees);
+
         // Withdraw from the account
-        require(IERC20(tracerQuoteToken).transfer(feeReceiver, tempFees), "TCR: Transfer failed");
-        emit FeeWithdrawn(feeReceiver, tempFees);
+        require(IERC20(tracerQuoteToken).transfer(feeReceiver, rawTokenFees), "TCR: Transfer failed");
+        emit FeeWithdrawn(feeReceiver, rawTokenFees);
```
kumar-ish, line 586: A comment or NatSpec should document that this is emitting an event with non-WAD values unlike the other events would be good

kumar-ish (approved):
:surfer:

PR #221, opened 2021-07-29
Validate orders in Trader.sol

OsmanBran:
# Motivation
https://github.com/code-423n4/2021-06-tracer-findings/issues/126

# Changes
- Validate that the matchOrders function returns true before processing a trade in trader.sol

### DIFF
--- a/contracts/Trader.sol
+++ b/contracts/Trader.sol
@@ -103,7 +103,7 @@ contract Trader is ITrader {
             // make low level call to catch revert
             // todo this could be succeptible to re-entrancy as
             // market is never verified
-            (bool success, ) = makeOrder.market.call(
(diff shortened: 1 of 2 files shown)


CalabashSquash (approved):

kumar-ish (commented):
I wonder if we should validate Tracer markets against the factory, if we want the trader to be a single source of truth (i.e. check the given market is in`validTracers` in `TracerPerpetualsFactory.sol`); this would ensure that in case a false market passed by a user (that isn't checked by the OME, due to a bug or otherwise) doesn't improperly impact the record-keeping of the trader contract. I think it'd be worth getting some opinions from @BenjaminPatch and @raymogg on this

OsmanBran:
Will address comment in future PR

PR #225, opened 2021-08-03
fix: Slippage protection patch

OsmanBran:
# Motivation
Address https://github.com/code-423n4/2021-06-tracer-findings/issues/105
Duplicate orders are not checked for in the slippage reimbursement function `calcUnitsSold`.

Following this patch, the system has the following restrictions:
- A receipt can only be claimed once. A receipt is still considered claimed even if the orders provided are invalid. therefore the orders provided must be checked carefully off-chain.
- Slippage claiming will revert if duplicate orders are provided.
- Slippage claiming will fail if the amount of the order was greater than the liquidation amount. I.e user liquidates 5 units, then sells 10 units at a loss. The user cannot claim slippage.

# Changes
- Add a mapping that tracks all **valid** processed orders and prevents duplication. Duplicate invalid orders are allowed.
- Note: an order is not valid for slippage reimbursement if it is: invalid trade or, valid trade but didn't result in any loss.
- Updated NatSpec to reflect above restrictions

### DIFF
--- a/contracts/Liquidation.sol
+++ b/contracts/Liquidation.sol
@@ -32,6 +32,9 @@ contract Liquidation is ILiquidation, Ownable {
     // Receipt ID => LiquidationReceipt
     mapping(uint256 => LibLiquidation.LiquidationReceipt) public liquidationReceipts;
 
+    // Order ID => used to claim slippage reimbursement
+    mapping(bytes32 => bool) claimedOrders;
+
     event ClaimedReceipts(address indexed liquidator, address indexed market, uint256 indexed receiptId);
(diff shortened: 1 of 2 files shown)


CalabashSquash (approved):

raymogg (approved):

raymogg:
On point 3 in the description, this means that we should ensure the liquidator bot doesn't try batch orders together right? since if it sells more than the receipt amount it will not be able to claim

OsmanBran:
That's right, additionally, even if that condition was relaxed, it would still fail if they batch together orders since an order can only be used to claim a receipt once.

PR #226, opened 2021-08-03
Use correct token format in drain pool and refactor

OsmanBran:
# Motivation
Address https://github.com/code-423n4/2021-06-tracer-findings/issues/99

# Changes
- Convert back to raw token value before calling aprove
- Refactor so that quote token address and decimals are saved as state variables


kumar-ish (approved):
Very nice. As a note for the future, I think we might want to consider renaming all instances of `collateralAsset` to quote tokens or something for consistency

PR #227, opened 2021-08-04
Fix hour tracking in Pricing

OsmanBran:
# Motivation
24 hour periods can go out of sync since `startLastHour`, used to determine if an hour elapsed was being updated to the current time whenever a new price was recorded.

For example:
Price recorded at hour 0, then 1.5 hours passes with no trades. New trade occurs so new price recorded, `startLastHour` is set to hour 1.5, but hour array is still at hour 1. Thus, the hourly recordings are now 0.5 hours out of sync. Can continue until the hours are significantly out of sync with 24 hours if there is very low trade volume.

Current hour increment also had the risk of integer overflow when elapsed hours exceeded 256 hours.

# Changes
- Correctly update `startLastHour` so that each 24 hour period is strictly 24 hours
- Refactor the updatePricing function
- Cast current hour to uint256 to prevent overflow

### DIFF
--- a/contracts/Pricing.sol
+++ b/contracts/Pricing.sol
@@ -71,41 +71,41 @@ contract Pricing is IPricing {
     /**
      * @notice Updates pricing information given a trade of a certain volume at
      *         a set price
-     * @param tradePrice the price the trade executed at
-     * @param fillAmount the amount the trade was filled for
+     * @param tradePrice The price the trade executed at
+     * @param fillAmount The amount the trade was filled for
      */
     function recordTrade(uint256 tradePrice, uint256 fillAmount) external override onlyTracer {
         uint256 currentOraclePrice = oracle.latestAnswer();
-        // update pricing information if a trade has not been recorded in the last hour
+        // Update pricing information if a trade has not been recorded in the last hour
         if (startLastHour <= block.timestamp - 1 hours) {
-            // get the last recorded hourly price, returns max integer if no trades occurred
+            // Get the last recorded hourly price, returns max integer if no trades occurred
             uint256 hourlyTracerPrice = getHourlyAvgTracerPrice(currentHour);
 
-            // emit the hourly average and udpate funding rate if trades occurred
+            // Emit the hourly average and udpate funding rate if trades occurred
             if (hourlyTracerPrice != type(uint256).max) {
-                // emit the old hourly average
+                // Emit the old hourly average
                 emit HourlyPriceUpdated(hourlyTracerPrice, currentHour);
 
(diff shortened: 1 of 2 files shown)


contracts/Pricing.sol
```diff
@@ -111,12 +111,12 @@ contract Pricing is IPricing {
...
                         staleHour = 23;
                     }
-                    updatePrice(0, 0, 0, true, staleHour);
+                    createPriceEntry(0, 0, 0, staleHour);
                 }
             }
 
             // update time of last hourly recording
-            startLastHour = block.timestamp;
+            startLastHour += (elapsedHours * 3600);
```
kumar-ish, line 119:
```suggestion
            startLastHour += (elapsedHours * 1 hours);
```

(I think this works)

contracts/Pricing.sol
```diff
@@ -125,54 +125,60 @@ contract Pricing is IPricing {
...
-                hourlyTracerPrices[hour].cumulativePrice +
-                PRBMathUD60x18.mul(marketPrice, fillAmount);
-            hourlyTracerPrices[hour].trades = hourlyTracerPrices[hour].trades + fillAmount;
-            // As above but with oracle price
-            hourlyOraclePrices[hour].cumulativePrice =
-                hourlyOraclePrices[hour].cumulativePrice +
-                PRBMathUD60x18.mul(oraclePrice, fillAmount);
-            hourlyOraclePrices[hour].trades = hourlyOraclePrices[hour].trades + fillAmount;
-        }
+        // Make new hourly record, total = marketprice, numTrades set to the amount filled;
```
kumar-ish:
```suggestion
        // Make new hourly record, total = marketPrice, numTrades set to the amount filled;
```

kumar-ish (changes requested):
Nice refactor on the `createPriceEntry`, that's been bugging me for months

contracts/Pricing.sol
OsmanBran: Not sure if this is makes it clearer. I personally think the constant 3600 makes it clearer that a conversion between hours to seconds is occurring.

contracts/Pricing.sol
OsmanBran: Fixed

contracts/Pricing.sol
kumar-ish: Is it possible to do the following?
```
            startLastHour += elapsedHours hours;
```

OsmanBran: Just tried, does not work

kumar-ish (approved):

PR #228, opened 2021-08-09
Added failure reason in failed order event

OsmanBran:
# Motivation
When the Tracer contract can't match orders it simply returns a `FailedOrders` event. This event does not provide any reasoning as to why the order fails and makes debugging difficult.

The event was updated to also return an error code stating the reason.

# Changes
- Add `OrderMatchingStatus` enum to list possible order failure reasons
- Make event emit error status depending on reason
- Refactor order validation in `matchOrders` function
- Add unit tests

### DIFF
--- a/contracts/lib/LibPerpetuals.sol
+++ b/contracts/lib/LibPerpetuals.sol
@@ -119,37 +119,6 @@ library Perpetuals {
     }
 
     /**
-     * @notice Checks if two orders can be matched given their price, side of trade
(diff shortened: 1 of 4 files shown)


contracts/lib/LibPerpetuals.sol
```diff
@@ -11,6 +11,19 @@ library Perpetuals {
         Short
     }
 
+    enum OrderMatchingResult {
```
OsmanBran, line 14: Docs will need to be updated at some point to explain error codes

CalabashSquash (approved):

PR #229, opened 2021-08-12
fix: Discard stale time values

OsmanBran:
# Motivation
The time value should return the average daily difference between the oracle and tracer price from the last 90 days.
Currently, the time value is incremented with a new daily difference (weighted to 90 days) every 24 hours. However, the time value is never decremented.

As a result, as soon as 90 days passes in the system, the time value will begin to deviate from the actual 90 day average due to the inclusion of older time values.

# Changes
- Create an array that stores daily price differences and update it each time a 24 hours passes
- Refactor calculateTWAP


CalabashSquash (approved):

raymogg (dismissed):

contracts/Pricing.sol
```diff
@@ -224,14 +229,25 @@ contract Pricing is IPricing {
...
+            int256 lastDailyDifference = Prices.timeValue(avgPrice, oracleAvgPrice);
+            timeValue += lastDailyDifference;
+            uint256 latestDay = lastUpdatedDay + elapsedDays;
+            for (uint256 currentDay = lastUpdatedDay + 1; currentDay <= latestDay; currentDay++) {
+                // add a new difference entry
+                int256 dailyDifference = (currentDay == latestDay) ? lastDailyDifference : int256(0);
+                dailyDifferences[currentDay] = dailyDifference;
+                // remove the difference entry 90 days ago
+                uint256 ninetyDaysAgo = currentDay - 90;
+                timeValue -= dailyDifferences[ninetyDaysAgo];
```
raymogg: I get what you are doing here, just wonder if there is a simpler way to do this in Solidity? The buffer is only ever 90 days long right? Could we just have an array with 90 days in it and simply make it circular? That way you just read from the index, subtract the value, then replace with the new daily difference.

Not sure if this would add much more complexity than its worth but just a thought.

raymogg (approved):

PR #230, opened 2021-08-13
feat: Remove token decimals as user input from constructor

OsmanBran:
# Motivation
Use ERC20.decimals() to get token decimals rather than relying on user input when setting the quote token decimals in the system.

# Changes
- Add IERC20Details.sol, interface that allows ERC20.decimals() to be accessed.
- Rely on ERC20.decimals() for quote token decimals in system.
- Add check that decimals is <= 18.


raymogg (approved):
Worth noting that this means all tokens used by the system MUST implement the `decimals()` function. Nothing major but we just need to be aware of that.

OsmanBran:
The `decimals()` function is implemented on all ERC20 tokens and returns 18 by default. So issues should only arise with the above when the token creator has incorrectly set up their ERC20 token / decimals (which should rarely occur).

PR #233, opened 2021-08-25
fix: Solidity coverage support

OsmanBran:
# Motivation
Solidity coverage is throwing a stack too deep error on current contracts.
This is an open issue to fix on the main solidity coverage repo. However, there is a fork of solidity coverage which has fixed this issue which can be used in the short-term. https://www.npmjs.com/package/@float-capital/solidity-coverage.

This change will be reverted after issue is resolved on main solidity coverage repo.

# Changes
- use forked version of solidity coverage


kumar-ish (approved):
:surfer:

PR #239, opened 2021-09-22
refactor: Insurance tests

OsmanBran:
# Motivation
Clean up Insurance tests. Remove smockit and replace with mock tracer contract.

# Changes
- Remove smockit, library can cause unintended behaviour with tests if not implemented carefully. Better to just use mock contract for simplicity and readability.
- Move common functions in tests to InsuranceUtil file
- Clean up set up for tests
- Remove unused trader unit tests


test/unit/InsuranceFundingRate.js
```diff
@@ -1,149 +1,49 @@
...
+    let quoteToken
+    let tracer
     let insurance
 
     beforeEach(async function () {
-        const _setup = await setup()
-        testToken = _setup.testToken
-        mockTracer = _setup.mockTracer
-        insurance = _setup.insurance
+        ;({ quoteToken, tracer, insurance } = await setupTests())
```
raymogg, line 38: This is some funky formatting, is this intentional?

OsmanBran: So here I'm directly destructuring the object returned by setupTests(). This allows us to skip initialising the `_setup` constant.

JS requires direct destructuring assignments to be wrapped in parentheses. The semicolon is required by the linter as it prevents the line accidently triggering a function in a previous line.

test/unit/InsuranceStaking.js
```diff
@@ -125,20 +56,17 @@ describe("Unit tests: Insurance.sol", function () {
                 expect(collateralHolding).to.equal(ethers.utils.parseEther("1"))
             })
 
-            it("pulls in collateral from the tracer market", async () => {
-                let balanceCalls = mockTracer.smocked.getBalance.calls.length
-                expect(balanceCalls).to.equal(1)
-            })
+            it("pulls in collateral from the tracer market", async () => {})
```
chrdevmar, line 59: is this test moved or was it meant to be deleted?

test/unit/InsuranceStaking.js
```diff
@@ -179,10 +107,7 @@ describe("Unit tests: Insurance.sol", function () {
...
```
chrdevmar, line 110: same with this one

test/unit/Trader.js
```diff
@@ -4,62 +4,17 @@ const { deploy } = deployments
...
-        perpMockAddress = await deployments.get("TracerPerpetualSwapMock")
-
-        let traderDeployment = await deployments.get("Trader")
-        trader = await ethers.getContractAt(
-            traderDeployment.abi,
-            traderDeployment.address
-        )
-        accounts = await ethers.getSigners()
-    })
+    before(async function () {})
```
chrdevmar, line 7: if theres no setup required might as well remove this

test/unit/Trader.js
```diff
@@ -4,62 +4,17 @@ const { deploy } = deployments
...
-            })
+            it("reverts", async () => {})
         })
         context("When the maker and taker array lengths differ", async () => {
-            it("reverts", async () => {
-                await expect(trader.executeTrade([], [])).to.be.revertedWith(
-                    "TDR: Received empty arrays"
-                )
-            })
+            it("reverts", async () => {})
```
chrdevmar, line 17: what are these about?

chrdevmar (changes requested):
the refactoring looks good, but there are a few test cases that now have empty bodies

test/unit/InsuranceFundingRate.js
raymogg: Interesting, for some reason I was under the impression you could just use `{ a, b } = someFunc()`. This is really cool though and good info, thanks

chrdevmar: one of the reasons im bullish on semicolons required in linting rules

OsmanBran:
In regards to the tests with removed bodies - these tests weren't testing behaviour thoroughly enough to be considered complete. As a result I have removed them and they will be added in a new PR so this one can be kept to purely refactoring.

test/unit/InsuranceFundingRate.js
chrdevmar: i believe you only need parentheses around destructuring if you are assigning to variables that are already initialised:

this is fine
```
const { a, b, c } = somefunc()
```

vs 

this needs them
```
let a;
let b;
let c;

({ a, b, c } = somefunc())
```

chrdevmar:
> In regards to the tests with removed bodies - these tests weren't testing behaviour thoroughly enough to be considered complete. As a result I have removed them and they will be added in a new PR so this one can be kept to purely refactoring.

sounds fair enough to me, can we maybe put a `.skip` on them or something? just because I'm pretty sure they will show up as passing tests and will be easy to just forget about them forever

OsmanBran:
The next major change I'll be working on is ensuring there is full coverage for all components so there shouldn't be much of a risk in terms of forgetting test cases as they'll be reviewed very soon. The skip could still be added in the short term though for clarity.

chrdevmar:
> The next major change I'll be working on is ensuring there is full coverage for all components so there shouldn't be much of a risk in terms of forgetting test cases as they'll be reviewed very soon. The skip could still be added in the short term though for clarity.

alright I'll leave it up to you, happy to approve

chrdevmar (approved):

PR #242, opened 2021-09-30
feat: Pricing test coverage

OsmanBran:
# Motivation
Edge case for when recordTrade was called for the first time, and updateTimeValue is called during this call, was not covered (Pricing.sol L#239).
Following this test case, there is full coverage except for a getter function (getHourlyOracleAvgPrice).

# Changes
- Add the test case
- Update some comments


chrdevmar (approved):
lgtm, assuming the failing test in CI expected

PR #243, opened 2021-10-05
feat: Tracer test coverage

OsmanBran:
# Motivation
Need to fill in some missing tests

# Changes
- Test that account leverage updates when user deposits/withdraws
- Test tracer liquidate functions (updateAccountsOnLiquidation and updateAccountsOnClaim)
- Change insurance check in updateAccountsOnClaim to be at top of function instead of end


contracts/TracerPerpetualSwaps.sol
```diff
@@ -478,12 +478,15 @@ contract TracerPerpetualSwaps is ITracerPerpetualSwaps, Ownable {
         int256 amountToTakeFromInsurance
     ) external override onlyLiquidation {
         address insuranceAddr = address(insuranceContract);
+        require(
+            balances[insuranceAddr].position.quote - amountToTakeFromInsurance >= 0,
+            "TCR: Insurance not funded enough"
+        );
```
chrdevmar, line 484: should this throw an error or should we instead just empty the insurance fund of everything its got, even if it holds less than the specified amount?

test/unit/TracerLiquidate.js
```diff
@@ -0,0 +1,244 @@
...
+                    tracer.updateAccountsOnLiquidation(
+                        liquidator.address,
+                        liquidatee.address,
+                        one,
+                        one,
+                        one,
+                        one,
+                        one
+                    )
+                ).to.be.revertedWith("TCR: Sender not liquidation")
```
chrdevmar, line 45: `Sender not liquidator` as opposed to `Sender not liquidation`?

test/unit/TracerLiquidate.js
```diff
@@ -0,0 +1,244 @@
...
+
+                await expect(
+                    tracer.updateAccountsOnClaim(
+                        liquidator.address,
+                        one,
+                        liquidator.address,
+                        one,
+                        one
+                    )
+                ).to.be.revertedWith("TCR: Sender not liquidation")
```
chrdevmar, line 145: same about the wording here

contracts/TracerPerpetualSwaps.sol
OsmanBran: I think this should throw an error. In the current state, this error should actually never be called since the function which calls it in `Liquidation.sol` determines how much should be taken from insurance (depending on if there's a sufficient amount). I.e Liquidation contract is responsible for handling any insufficient funds not the tracer contract which is just there to update balances.

test/unit/TracerLiquidate.js
OsmanBran: These test cases check if `Liquidation.sol` has called the function, so this should be all good.

chrdevmar (approved):

PR #244, opened 2021-10-05
feat: Library test coverage

OsmanBran:
# Motivation
Add some missing test cases

# Changes
- Test for when deleveragingCliff == insurancePoolSwitchStage
- Test for when periods of no trades are included in calculateTWAP arrays


contracts/lib/LibPerpetuals.sol
```diff
@@ -86,9 +86,10 @@ library Perpetuals {
         uint256 deleveragingCliff,
         uint256 insurancePoolSwitchStage
     ) internal pure returns (uint256) {
-        if (poolTarget == 0) {
+        if (deleveragingCliff == insurancePoolSwitchStage || poolTarget == 0) {
```
OsmanBran: This case will never occur in our protocol since deleveragingCliff must be set higher than insurancePoolSwitchStage.

However, it has been included as a special case in this function in case other implementations of tracer do not have the same requirement. It might be better to change this to a require() instead of special case.

chrdevmar: should this be `deleveragingCliff >= insurancePoolSwitchStage`?

OsmanBran: Good catch, yes it should be. Although should be noted that `deleveragingCliff` should always be less than `insurancePoolSwitchStage` by definition. So now I think about it it probably makes more sense to have it as a require() check.

chrdevmar: ok sounds good

chrdevmar (changes requested):
changes look good, getting a `MODULE_NOT_FOUND` error in CI though

chrdevmar (approved):
lgtm

